Issue #8, opened 2017-04-10
Remove ovirt inventory code from manageiq-gems-pending


Fryguy:
Now that the oVirt provider has its own repo, it's time to move the ovirt specific code from manageiq-gems-pending to here. See https://github.com/ManageIQ/manageiq-gems-pending/tree/master/lib/gems/pending/ovirt_provider.

jhernand:
@miq-bot assign @jhernand

bdunne:
Both PR's are merged, so I closed the issue

PR #9, opened 2017-04-11
Move code from manageiq-gems-pending

jhernand:
Now that the oVirt provider has been extracted to a separate repository,
it makes sense to move all the oVirt specific code from
'manageiq-gems-pending' to this repository.

This patch moves the 'ovirt_event_monitor.rb' and 'ovirt_inventory.rb'
files. These files implement event monitoring and inventory collection,
using version 3 of the oVirt API and the 'ovirt' gem. As there is an
ongoing effort to use version 4 of the oVirt API, and the oVirt SDK,
these files have been added to a new 'ManageIQ::Providers::Ovirt::Legacy'
module, to make it explicit that this code will eventually be removed.

https://github.com/ManageIQ/manageiq-providers-ovirt/issues/8

### DIFF
--- a/lib/manageiq/providers/ovirt/legacy/event_monitor.rb
+++ b/lib/manageiq/providers/ovirt/legacy/event_monitor.rb
@@ -1,0 +1,50 @@
+require 'ovirt'
+require 'manageiq/providers/ovirt/legacy/inventory'
+
+module ManageIQ
+  module Providers
+    module Ovirt
+      module Legacy
+        class EventMonitor
+          def initialize(options = {})
+            @options = options
+          end
+
+          def inventory
+            ::Ovirt.logger = $rhevm_log if $rhevm_log
+
+            @inventory ||= ManageIQ::Providers::Ovirt::Legacy::Inventory.new(@options)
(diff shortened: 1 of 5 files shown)


jhernand:
Note that this should be merged before ManageIQ/manageiq-gems-pending#118.

lib/manageiq/providers/ovirt/legacy/event_monitor.rb
```diff
@@ -0,0 +1,50 @@
+require 'ovirt'
```
durandom: Any reason you chose to move them to `lib/`?
I think this all could easily go to `app/models/manageiq/providers/redhat`

jhernand: The only reason is that these classes are used by the those test scripts that I don't know where to place. Those are in the `manageiq-gems-pending` repo. Not sure if they can be used from that repo if they aren't in the `lib` directory of this gem. Once we decide what we do with those test scripts, an once I know how to test them, I can move the files to `app/models` no objection to that.

PR #14, opened 2017-04-19
Event parser tests

pkliczewski:
We need addtional tests for EventParser#parse_new_target

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
@@ -78,7 +78,11 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParser
   end
 
   def self.parse_target_name(message, event_type)
-    if %w(NETWORK_ADD_VM_INTERFACE NETWORK_INTERFACE_PLUGGED_INTO_VM).include?(event_type)
+    case event_type
+    when "NETWORK_ADD_VM_INTERFACE"
+      # sample message: "Interface nic1 (VirtIO) was added to VM v5. (User: admin@internal-authz)"
+      message.split(/\s/)[7][0...-1]
+    when "NETWORK_INTERFACE_PLUGGED_INTO_VM"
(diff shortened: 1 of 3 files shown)


agrare:
Hey @pkliczewski looks great!  I'd like to try to include testing the result with `EmsRefresh.refresh_new_target` to make sure that any change in that interface would be caught.

pkliczewski:
@miq-bot assign @durandom

pkliczewski:
@masayag @borod108 please take a look

durandom:
@borod108 @masayag could you approve or req changes to this PR?

app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
```diff
@@ -78,7 +78,11 @@ def self.parse_new_vm(vm, message, event_type, ems)
   end
 
   def self.parse_target_name(message, event_type)
-    if %w(NETWORK_ADD_VM_INTERFACE NETWORK_INTERFACE_PLUGGED_INTO_VM).include?(event_type)
+    case event_type
+    when "NETWORK_ADD_VM_INTERFACE"
+      # sample message: "Interface nic1 (VirtIO) was added to VM v5. (User: admin@internal-authz)"
+      message.split(/\s/)[7][0...-1]
```
pkliczewski: It is a limitation of a model that we use and there are edge cases where it is the only way to understand the name. I agree that we rely on fragile message but there is no other way around it.

masayag: Ok, so since this requires some work out of the scope of this PR, I'd suggest to approve this one which solves a bug around the parsing of NETWORK_ADD_VM_INTERFACE and to open a bug for replacing the logic to identify the targets based on their ID once it is possible instead.

pkliczewski: We have new card to tack it https://trello.com/c/L9rtqKGw/140-improve-event-parsing-for-refresh-for-new-target

PR #21, opened 2017-04-26
Add support for v2v async task

matobet:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1404920

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
@@ -30,6 +30,13 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
     )
   end
 
+  def submit_import_vm(userid, source_vm_id, target_params)
+    task_id = queue_import_vm(userid, source_vm_id, target_params)
+    task = MiqTask.wait_for_taskid(task_id)
+
+    task.task_results
+  end
+
   def validate_import_vm
     highest_supported_api_version && highest_supported_api_version >= '4'
   end
@@ -45,9 +52,31 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
     end
   end
 
+  def queue_import_vm(userid, source_vm_id, target_params)
+    task_options = {
+      :action => 'Import VM',
+      :userid => userid
+    }
+
+    queue_options = {
+      :task_id     => nil, # run this task concurrently,
+      # since this is called by an (automate) task without this the task_id would be inherited from it
+      # and cause a deadlock in subsequent waiting on this task from synchronous context
+      :zone        => my_zone,
+      :class_name  => self.class.name,
+      :method_name => 'import_vm',
+      :instance_id => id,
+      :role        => 'ems_operations',
+      :args        => [source_vm_id, target_params]
+    }
+
+    _log.info("Queueing import of VM ID: #{source_vm_id} by user #{userid}")
+    MiqTask.generic_action_with_callback(task_options, queue_options)
+  end
+
   def perform_vmware_to_ovirt_import(params)
     with_provider_connection :version => 4 do |conn|
-      conn.system_service.external_vm_imports_service.add(
+      import = conn.system_service.external_vm_imports_service.add(
         OvirtSDK4::ExternalVmImport.new(
           :name           => params[:source_vm_name],
           :vm             => { :name => params[:target_vm_name] || params[:source_vm_name] },
@@ -60,6 +89,7 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
           :sparse         => params[:sparse],
         )
       )
+      self.class.make_ems_ref(import.vm.href)
     end
   end
 
--- a/content/service_dialogs/transform-vm.yml
+++ b/content/service_dialogs/transform-vm.yml
@@ -1,0 +1,217 @@
+---
+- description: Transform VM
+  buttons: submit,cancel
+  label: Transform VM
+  blueprint_id:
+  dialog_tabs:
+  - description:
+    display: edit
+    label: Main
+    display_method:
+    display_method_options:
+    position: 0
+    dialog_groups:
+    - description: Import given virtual machine to another infrastructure provider
+      display: edit
+      label: Transform Virtual Machine
+      display_method:
+      display_method_options:
+      position: 0
+      dialog_fields:
+      - name: name
+        description: Name of the newly created virtual machine
+        data_type:
+        notes:
+        notes_display:
+        display: edit
+        display_method:
+        display_method_options: {}
+        required: true
+        required_method:
+        required_method_options: {}
+        default_value: ''
+        values:
+        values_method:
+        values_method_options: {}
+        options:
+          :protected: false
+        label: Name
+        position: 0
+        validator_type:
+        validator_rule:
+        reconfigurable:
+        dynamic: false
+        show_refresh_button:
+        load_values_on_init:
+        read_only:
+        auto_refresh:
+        trigger_auto_refresh:
+        visible: true
+        type: DialogFieldTextBox
+        resource_action:
+          action:
+          resource_type: DialogField
+          ae_namespace:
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -60,9 +84,14 @@ def perform_vmware_to_ovirt_import(params)
           :sparse         => params[:sparse],
         )
       )
+      ems_ref_for_vm_id(import.vm.id)
     end
   end
 
+  def ems_ref_for_vm_id(vm_id)
+    "/api/vms/#{vm_id}"
```
mkanoor: @matobet Is this how RHEVM's represents the ems_ref ids and how we store in our database? Does this string match how the ems_refresh is going to build the vm information

matobet: @mkanoor this format is taken directly from the database and I imagine the refresh is using exactly those REST paths to reference RHEVM's entities

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -45,9 +52,26 @@ def check_import_supported!(source_provider)
...
+    task_options = {
+      :action => 'Import VM',
+      :userid => userid
+    }
+
+    queue_options = {
+      :class_name  => self.class.name,
+      :method_name => 'import_vm',
+      :instance_id => self.id,
+      :role        => 'ems_operation',
```
durandom: an `s` is missing here

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -45,9 +53,28 @@ def check_import_supported!(source_provider)
...
+      :userid => userid
+    }
+
+    queue_options = {
+      :zone        => self.my_zone,
+      :class_name  => self.class.name,
+      :method_name => 'import_vm',
+      :instance_id => self.id,
+      :role        => 'ems_operations',
+      :args        => [source_vm_id, target_params]
```
mkanoor: @matobet 
Can you add :task_id = nil here for the work to be picked up right away. If you dont have a nil it uses the previous task_id and will block

matobet:
@miq-bot assign @borod108

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -60,9 +89,14 @@ def perform_vmware_to_ovirt_import(params)
           :sparse         => params[:sparse],
         )
       )
+      ems_ref_for_vm_id(import.vm.id)
```
borod108: any chance inport.vm or vm.id will be nill if there was no exception?

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
borod108: @matobet I think you should get it from import.vm.href but that one has /ovirt-engine/ in the beginning so it should be removed. I think it is more stable to remove the prefix than to build the entire ref in case of future changes. 
@jhernand waht is the best approach in this case?

jhernand: We should always avoid building paths like this. We had (and will have) many issues because of changing or incorrect paths. As in the previous step you already got a VM object, which should have an `href` attribute, the right way to calculate the EMS reference is to use the `make_ems_ref` method, defined here:

https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb#L340-L345

That `make_ems_ref` should be the only place where we do this kind of path manipulation.

To use it you need something like this:

```ruby
ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(import.vm.href)
```

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
matobet: I don't believe so, If the API returns an 20x OK status it should contain the basic vm ID data

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
matobet: @jhernand done

matobet:
@miq-bot add_label fine/yes

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -60,6 +89,7 @@ def perform_vmware_to_ovirt_import(params)
           :sparse         => params[:sparse],
         )
       )
+      make_ems_ref(import.vm.href)
```
borod108: according to the specs this does not work.

matobet: yes the tests are indeed failing, but it works when deployed and run within the EVM.... could it be possible that the tests just don't see methods defined in the `api_integration` mixin? @borod108 @jhernand

matobet: Ok it seems a small typo from my side, the tests are passing now :)

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -60,6 +89,7 @@ def perform_vmware_to_ovirt_import(params)
           :sparse         => params[:sparse],
         )
       )
+      self.class.make_ems_ref(import.vm.href)
```
jhernand: I know that the fully qualified name of this method is very long, but I'd say that assuming that it is available in this class isn't correct. Besides, if this code is moved to a different place, it will not work. I think it would be better to use the ugly fully qualified name, but that is probably a matter of taste. So +0.99 from me.

matobet:
@miq-bot assign @durandom

content/service_dialogs/transform-vm.yml
```diff
@@ -0,0 +1,221 @@
...
+          ae_instance:
+          ae_message:
+          ae_attributes: {}
+  resource_actions:
+    - action:
+      ae_namespace: SYSTEM
+      ae_class: PROCESS
+      ae_instance: Request
+      ae_message:
+      ae_attributes:
```
tinaafitz: @matobet Can you create a /ManageIQ/System/Request/importvm instance instead of using call_instance here?  Call_instance is intended to be used for Simulate and testing.

matobet: @tinaafitz please forgive my ignorance of the automation method invocation syntax, but how exactly would that look like?

tinaafitz: @matobet, Automate can be really confusing, I should have added more detail.  The solution would be to change the request value below to importvm, or some new instance name of your choosing.  You can remove the values for namespace, class, and instance here. 
You would add a new instance in the ManageIQ-content repo for /ManageIQ/System/Request class(importvm or your new name) and you would specify the relationship there to point directly to: ManageIQ/Infrastructure/VM/Transform/Import/ImportVm. Look at 
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/ansible_tower_job.yaml as an example.

matobet: @tinaafitz thank you very much for your explanation! Posted new version.

oourfali:
@durandom right.

simaishi:

```
$ git log -1
commit 9cf503b1ada06d1e528667a321fd6dabdb18f42f
Date:   Thu May 18 12:58:52 2017 +0300

    Merge pull request #21 from matobet/v2v-queue
    
    Add support for v2v async task
    (cherry picked from commit dfb1d9246c2f42c8e5bbbbd960b4c728e49d92fe)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459996
```

PR #23, opened 2017-04-27
Preparation for graph refresh implementation

pkliczewski:
RFE link: https://bugzilla.redhat.com/1466172

### DIFF
--- a/app/models/manageiq/providers/redhat/builder.rb
+++ b/app/models/manageiq/providers/redhat/builder.rb
@@ -1,0 +1,45 @@
+class ManageIQ::Providers::Redhat::Builder
+  class << self
+    def build_inventory(ems, target)
+      if target.kind_of? ManagerRefresh::TargetCollection
+        inventory(
+          ems,
+          target,
+          ManageIQ::Providers::Redhat::Inventory::Collector::TargetCollection,
+          ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection,
+          [parser]
+        )
+      else
+        # Fallback to ems refresh or full refresh
+        infra_manager_inventory(ems, target)
+      end
+    end
+
+    private
+
+    def parser
+      ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager
+    end
+
+    def infra_manager_inventory(ems, target)
+      inventory(
+        ems,
+        target,
+        ManageIQ::Providers::Redhat::Inventory::Collector::InfraManager,
+        ManageIQ::Providers::Redhat::Inventory::Persister::InfraManager,
+        [parser]
+      )
+    end
+
+    def inventory(manager, raw_target, collector_class, persister_class, parsers_classes)
+      collector = collector_class.new(manager, raw_target)
+      persister = persister_class.new(manager, raw_target, collector)
+
+      ::ManageIQ::Providers::Redhat::Inventory.new(
+        persister,
+        collector,
+        parsers_classes.map(&:new)
+      )
+    end
+  end
+end
--- a/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
@@ -6,20 +6,24 @@ module ManageIQ::Providers::Redhat::InfraManager::Refresh
       inventory = inventory_from_ovirt(ems)
       raise "Invalid RHEV server ip address." if inventory.api.nil?
 
-      # TODO before iterating over targets it would be good to check whether ExtMgmntSystem is part of it
+      # TODO: before iterating over targets it would be good to check whether ExtMgmntSystem is part of it
       # TODO optimize not to fetch the same objects like clusters for multiple targets
 
       targets_with_data = targets.collect do |target|
         _log.info "Filtering inventory for #{target.class} [#{target.name}] id: [#{target.id}]..."
 
-        case target
-        when Host
-          data,  = Benchmark.realtime_block(:fetch_host_data) { host_targeted_refresh(inventory, target) }
-        when VmOrTemplate
-          data,  = Benchmark.realtime_block(:fetch_vm_data) { vm_targeted_refresh(inventory, target) }
+        # TODO: make sure not to use with api v3
+        if refresher_options.try(:[], :inventory_object_refresh)
+          data = ManageIQ::Providers::Redhat::Builder.build_inventory(ems, target)
         else
-          data,  = Benchmark.realtime_block(:fetch_all) { inventory.refresh }
-
+          case target
+          when Host
+            data,  = Benchmark.realtime_block(:fetch_host_data) { host_targeted_refresh(inventory, target) }
+          when VmOrTemplate
+            data,  = Benchmark.realtime_block(:fetch_vm_data) { vm_targeted_refresh(inventory, target) }
+          else
+            data,  = Benchmark.realtime_block(:fetch_all) { inventory.refresh }
+          end
         end
 
         _log.info "Filtering inventory...Complete"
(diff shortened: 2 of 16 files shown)


pkliczewski:
@Ladas @borod108 @masayag Please take a look

pkliczewski:
[Graph refresh](https://trello.com/c/n8COkeES/139-graph-refresh)

app/models/manageiq/providers/redhat/infra_manager/refresh/strategies/api3.rb
```diff
@@ -1,5 +1,9 @@
 module ManageIQ::Providers::Redhat::InfraManager::Refresh::Strategies
   class Api3 < ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher
+    def self.ems_type
+      @ems_type ||= "rhevm".freeze
```
Ladas: hm, why do we need this one? should be on the Manager object

app/models/manageiq/providers/redhat/inventory/collector/target_collection.rb
```diff
@@ -0,0 +1,87 @@
+class ManageIQ::Providers::Redhat::Inventory::Collector::TargetCollection < ManageIQ::Providers::Redhat::Inventory::Collector
+  def initialize(_manager, _target)
+    super
+    parse_targets!
+    infer_related_ems_refs!
+
+    # Reset the target cache, so we can access new targets inside
+    target.manager_refs_by_association_reset
```
Ladas: gah, this is my ugliness, I plan to refactor it to something nicer :-)

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
```diff
@@ -0,0 +1,75 @@
...
+    # TODO: check what needs to be added here
+  end
+
+  def references(collection)
+    target.manager_refs_by_association.try(:[], collection).try(:[], :ems_ref).try(:to_a) || []
+  end
+
+  # TODO: check whether needed
+  def name_references(collection)
+    target.manager_refs_by_association.try(:[], collection).try(:[], :name).try(:to_a) || []
```
Ladas: I need to refactor these, it's basically just about extracting IDs of some kinds from all the targets we have, the current structure is not nice

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
```diff
@@ -0,0 +1,75 @@
...
+      )
+    )
+  end
+
+  # TODO: check correctness
+  def add_miq_templates_inventory_collections(manager_refs)
+    return if manager_refs.blank?
+
+    add_inventory_collection(
+      infra.miq_templates(
```
Ladas: can you turn Template into Vm and vice versa (keeping the ems_ref) like VMWare? In that case it would make sense to save Vms and Templates together

app/models/manageiq/providers/redhat/inventory/parser.rb
```diff
@@ -0,0 +1,14 @@
+class ManageIQ::Providers::Redhat::Inventory::Parser < ManagerRefresh::Inventory::Parser
+  require_nested :InfraManager
+
+  def process_inventory_collection(collection, key)
```
Ladas: I think we will not need this one, using the new syntax of the refresh

Ladas:
few nits, otherwise this looks like a great start 👍

app/models/manageiq/providers/redhat/infra_manager/refresh/strategies/api3.rb
pkliczewski: It is needed due to our refresh structure using strategies. We call `parent.ems_type` which in our case is `ManageIQ::Providers::Redhat::InfraManager::Refresh::Strategies` which do not have required method.

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
pkliczewski: I was not sure whether I should use here `VmOrTemplate` or `Vm` and `Template`. It should be clarified once I start to work on the oVirt specific code.

app/models/manageiq/providers/redhat/inventory/parser.rb
pkliczewski: Please point me to new syntax

Ladas: e.g. https://github.com/Ladas/manageiq-providers-amazon/blob/f8615fb4c69e7485b57f4f1ffd4f6bbbc4782438/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb#L169-L169

so process_inventory_collection is not needed anymore

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
Ladas: Sounds good.

app/models/manageiq/providers/redhat/infra_manager/refresh/strategies/api3.rb
Ladas: ok

pkliczewski:
@Ladas @borod108 @masayag Do you think we could merge this PR. It does not make any harm to the code. We need to split the work somehow.

Ladas:
@pkliczewski sure, it's a nice first step, so I would just merge. It's all hidden under the settings refresher_options.try(:[], :inventory_object_refresh), so it should not cause any failures.

👍

pkliczewski:
@miq-bot assign @oourfali

borod108:
@pkliczewski I wonder if we could make a namespace for all of the graph refresh?

pkliczewski:
@borod108 As far as I understand the file layout/namespacing is exactly how graph refresh is expecting.
@Ladas Am I right?

Ladas:
@pkliczewski @borod108 the current namespace is ManageIQ::Providers::Redhat::Inventory::, but you can have another level if you wish. Is that what you are after? :-)

pkliczewski:
@Ladas @borod108 had some concerns since we are using inventory name a lot. Thank you!

pkliczewski:
[Graph refresh](https://trello.com/c/uid9QN8b/161-graph-refresh)

app/models/manageiq/providers/redhat/builder.rb
```diff
@@ -0,0 +1,48 @@
+class ManageIQ::Providers::Redhat::Builder
+  class << self
+    def build_inventory(ems, target)
+      case target
+      when ManageIQ::Providers::Redhat::InfraManager
```
masayag: the switch can be replace by a simple if-else statement:


`
if target == ManageIQ::Providers::Redhat::InfraManager
    inventory(...)
 else
    infra_manager_inventory(...)
end
`

app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
```diff
@@ -6,20 +6,24 @@ def collect_inventory_for_targets(ems, targets)
...
       targets_with_data = targets.collect do |target|
         _log.info "Filtering inventory for #{target.class} [#{target.name}] id: [#{target.id}]..."
 
-        case target
-        when Host
-          data,  = Benchmark.realtime_block(:fetch_host_data) { host_targeted_refresh(inventory, target) }
-        when VmOrTemplate
-          data,  = Benchmark.realtime_block(:fetch_vm_data) { vm_targeted_refresh(inventory, target) }
+        # TODO: make sure not to use with api v3
+        if refresher_options.try(:[], :inventory_object_refresh)
```
borod108: can you please explain this check? I see it repeats several times. Is this a way to determine whether graph refresh is used?

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_target_vm_4_spec.rb
```diff
@@ -37,40 +37,54 @@ def load_response_mock_for(filename)
...
-    assert_cluster(vm)
-    assert_storage(storage, vm)
+    assert_vm(@vm, @storage)
+    assert_vm_rels(@vm, @hardware, @storage)
+    assert_cluster(@vm)
+    assert_storage(@storage, @vm)
+  end
+
+  it "should collect a vm" do
+    allow(Settings.ems_refresh).to receive(:rhevm).and_return(:inventory_object_refresh => true)
```
borod108: just for consistency, in other tests I used something like:
stub_settings_merge(:ems => { :ems_redhat => { :use_ovirt_engine_sdk => use_ovirt_engine_sdk } })

Not sure its better, maybe actually the other way around. I just think consistency is nice (but not mandatory here)

app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
pkliczewski: This check determines whether the graph refresh is used or regular refresh. Similar approach we have for other providers.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_target_vm_4_spec.rb
pkliczewski: I agree with consistency issue and will fix.

app/models/manageiq/providers/redhat/builder.rb
pkliczewski: It could be done as you suggest. This piece is based on the logic from other providers.
case graph_full:
case graph_target:
else regualr_refresh

To keep this clarity we would need to have if elseif and else which is not that different.

app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
borod108: is it a hash? if so, maybe "has_key?" will do?

pkliczewski: What do you think about being consistent with other providers like [1]?


[1] https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb#L10

borod108: I think it does not look nice, any idea why they chose this way of doing it?
(I can only think of the case its not a hash and they actually do not know what kind of argument it will be, and even than I think there should be a nicer solution.)

pkliczewski: Not sure why it was done in this way.

borod108: @Ladas @durandom can you please explain why it was written this way? (is it the recommended style?)

pkliczewski: It is not a hash but `Config::Options` if value provided otherwise it is `nil`

PR #27, opened 2017-05-04
ovirt-networking: using profiles

AlonaKaplan:
This patch fixes some issues in oVirt vm provision network tab.
1. Changing the 'vLan' label to 'Network'
2. Displaying the vnic profiles instead of the host/hosts networks (only
profiles configured in the cluster).
3. Using OvirtSDK4.

Changes relevant to version 4 only.

Signed-off-by: Alona Kaplan <alkaplan@redhat.com>

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
@@ -8,15 +8,8 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration::Netw
       return
     end
 
-    requested_vnics.stretch!(destination_vnics).each_with_index do |requested_vnic, idx|
-      if requested_vnic.nil?
-        # Remove any unneeded vm nics
-        destination_vnics[idx].destroy
-      else
-        # Create or update existing nics
-        configure_vnic(requested_vnic, "nic#{idx + 1}", destination_vnics[idx])
-      end
-    end
+    ems = destination.ext_management_system
+    ems.ovirt_services.configure_vnics(requested_vnics, destination_vnics, dest_cluster, destination)
   end
 
   # TODO: Move this into EMS Refresh
@@ -57,21 +50,5 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration::Netw
       _log.info("vlan: #{vlan.inspect}")
       {:network => vlan, :mac_address => get_option_last(:mac_address)}
     end
-  end
-
-  def configure_vnic(network_hash, name, vnic)
-    mac_addr  = network_hash[:mac_address]
-    network   = find_network_in_cluster(network_hash[:network])
-    raise MiqException::MiqProvisionError, "Unable to find specified network: <#{network_hash[:network]}>" if network.nil?
-
-    ems.ovirt_services.configure_vnic(
-      :vm        => destination,
-      :mac_addr  => mac_addr,
-      :network   => network,
-      :nic_name  => name,
-      :interface => network_hash[:interface],
-      :vnic      => vnic,
-      :logger    => _log
-    )
   end
 end
--- a/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
@@ -92,4 +92,20 @@ class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisio
 
     result.select { |s| s.storage_domain_type == "data" }
   end
+
+  def source_ems
+    src = get_source_and_targets
+    load_ar_obj(src[:ems])
+  end
+
+  def load_allowed_vlans(hosts, vlans)
+    ems = source_ems
+    ems.ovirt_services.load_allowed_networks(hosts, vlans, self)
+  end
+
+  def filter_allowed_hosts(all_hosts)
+    ems = source_ems
+    ovirt_services = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Builder.new(ems).build(:use_highest_supported_version => true).new(:ems => ems)
+    ovirt_services.filter_allowed_hosts(self, all_hosts)
+  end
 end
--- a/spec/models/manageiq/providers/redhat/infra_manager/provision/configuration/network_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/provision/configuration/network_spec.rb
@@ -5,9 +5,6 @@ describe ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration::Ne
   let(:network_id)    { "network1-id" }
   let(:network_name)  { "network1-name" }
   let(:rhevm_cluster) { double("Ovirt::Cluster", :find_network_by_name => {:id => network_id}) }
-  let(:rhevm_nic1)    { {:name => "nic1", :network => {:id => network_id}, :mac => {:address => mac_address}} }
-  let(:rhevm_nic2)    { {:name => "nic2", :network => {:id => "network2-id"}} }
-  let(:set_vlan)      { @task.options[:vlan] = [network_name, network_name] }
   let(:ems)           { FactoryGirl.create(:ems_redhat_with_authentication) }
   let(:ems_cluster)   { FactoryGirl.create(:ems_cluster, :ext_management_system => ems) }
   let(:template)      { FactoryGirl.create(:template_redhat, :ext_management_system => ems) }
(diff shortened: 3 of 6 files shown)


AlonaKaplan:
@borod108 @masayag  please review
depends on https://github.com/ManageIQ/manageiq/pull/14991

AlonaKaplan:
@Ladas please review

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -80,4 +91,37 @@ def configure_vnic(network_hash, name, vnic)
...
+        if requested_vnic.nil?
+          nics_service.nic_service(destination_vnics[idx][:id]).remove
+        else
+          configure_v4_vnic(requested_vnic, create_new_vnic_name(idx), destination_vnics[idx], nics_service)
+        end
+      end
+    end
+  end
+
+  def configure_v4_vnic(network_hash, name, vnic, nics_service)
```
Ladas: nit: same order of parameters as configure_v3_vnic, with one extra?

borod108:
In general we have ovirt_service/strategies/v4 or v3 and I think configure_v4_vnics/configure_v3_vnics shuold go there and have the same params passed.

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -80,4 +91,37 @@ def configure_vnic(network_hash, name, vnic)
       :logger    => _log
     )
   end
+
+  def configure_v4_vnics(requested_vnics)
+    destination.ext_management_system.with_provider_connection(:version => 4) do |connection|
+      nics_service = connection.system_service.vms_service.vm_service(destination.uid_ems).nics_service
```
Ladas: I would probably pack it as vm.provider_nics_service to ManageIQ::Providers::Redhat::InfraManager::Vm, similar to provider_object method. And I wonder if it can leverage the provider_object method?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -92,4 +92,39 @@ def allowed_storages(options = {})
 
     result.select { |s| s.storage_domain_type == "data" }
   end
+
+  def load_allowed_vlans(hosts, vlans)
+    ems = source_ems
+    if update_vnic_profile_supported?(ems)
```
Ladas: you can probably just call ems.supports_update_vnic_profile? here, unless you plan to have more logic in update_vnic_profile_supported?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -92,4 +92,39 @@ def allowed_storages(options = {})
...
+    load_ar_obj(src[:ems])
+  end
+
+  def update_vnic_profile_supported?(ems)
+    ems.supports_update_vnic_profile?
+  end
+
+  def load_v4_allowed_vlans(ems, vlans)
+    ems.with_provider_connection(:version => 4) do |connection|
+      profiles = connection.system_service.vnic_profiles_service.list
```
Ladas: the connection logic should be probably in ManageIQ::Providers::Redhat::InfraManager::ApiIntegration

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -92,4 +92,39 @@ def allowed_storages(options = {})
...
+
+      uid_ems_cluster = VmOrTemplate.find(get_source_vm.id).ems_cluster.uid_ems
+      cluster_networks = connection.system_service.clusters_service.cluster_service(uid_ems_cluster).networks_service.list
```
Ladas: same for connection logic in ManageIQ::Providers::Redhat::InfraManager::ApiIntegration

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -92,4 +92,39 @@ def allowed_storages(options = {})
...
+    src = get_source_and_targets
+    load_ar_obj(src[:ems])
+  end
+
+  def update_vnic_profile_supported?(ems)
+    ems.supports_update_vnic_profile?
+  end
+
+  def load_v4_allowed_vlans(ems, vlans)
+    ems.with_provider_connection(:version => 4) do |connection|
```
Ladas: Also we don't usually go directly to the API here, but rather refresh the data and then show it from a MIQ DB. This code will run in UI worker, then the potential issues are, that it can be blocking the worker, or the appliance with UI worker is not configured for reaching the Provider.

@agrare do you think we can do this as a first version, before the refresh is done, or we tend not to do this, because of the issues I've mentioned and possibly others?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
masayag: `vnic_profiles_service.list` returns all of the vnic profiles in the system.
We should restrict the selection only for vnic profiles available in the context of the template's cluster.

@jhernand Is there an option to filter them on the rest api call by providing a filter to the SDK call ?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -92,4 +92,39 @@ def allowed_storages(options = {})
...
+      profiles = connection.system_service.vnic_profiles_service.list
+
+      uid_ems_cluster = VmOrTemplate.find(get_source_vm.id).ems_cluster.uid_ems
+      cluster_networks = connection.system_service.clusters_service.cluster_service(uid_ems_cluster).networks_service.list
+
+      profiles.each do |p|
```
masayag: The allowed vlan will contains all of the vnic profiles allowed for a vm in a specific cluster.
However, if the user selects a specific host to be run on, and some of the networks are defined as 'optional networks' on ovirt-engine. we should present only the networks which are currently available on the selected host.
Else, the vm will fail to run on that host.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
agrare: @Ladas yeah we had issues with this for dvswitches also, a lot of times the UI worker can't even route to the provider directly so it can fail to connect at all.

If we merge this as is we should create an issue to track needing to add this to vmdb so we don't forget about it

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
AlonaKaplan: Couldn't find a way to filter them on the rest call.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
AlonaKaplan: The logic here is same we have in ovirt (showing all the vnic profiles belong to vm networks in the cluster). I don't think we should have here more restrictions than ovirt has.

BTW, there are use case when we allow running a vm on a host that doesn't contain the non-rquired network (If ConfigValues.OnlyRequiredNetworksMandatoryForVdsSelection is true).

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
AlonaKaplan: done

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -8,14 +8,10 @@ def configure_network_adapters
...
-    requested_vnics.stretch!(destination_vnics).each_with_index do |requested_vnic, idx|
-      if requested_vnic.nil?
-        # Remove any unneeded vm nics
-        destination_vnics[idx].destroy
-      else
-        # Create or update existing nics
-        configure_vnic(requested_vnic, "nic#{idx + 1}", destination_vnics[idx])
-      end
+    if destination.ext_management_system.supports_update_vnic_profile?
+      configure_v4_vnics(requested_vnics)
```
AlonaKaplan: done

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -80,4 +91,37 @@ def configure_vnic(network_hash, name, vnic)
       :logger    => _log
     )
   end
+
+  def configure_v4_vnics(requested_vnics)
+    destination.ext_management_system.with_provider_connection(:version => 4) do |connection|
```
AlonaKaplan: done

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -80,4 +91,37 @@ def configure_vnic(network_hash, name, vnic)
...
+          nics_service.nic_service(destination_vnics[idx][:id]).remove
+        else
+          configure_v4_vnic(requested_vnic, create_new_vnic_name(idx), destination_vnics[idx], nics_service)
+        end
+      end
+    end
+  end
+
+  def configure_v4_vnic(network_hash, name, vnic, nics_service)
+    profile_id = network_hash[:network]
```
AlonaKaplan: done

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
AlonaKaplan: done

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
```diff
@@ -65,19 +59,67 @@ def configure_dialog_nic
...
+  #     :vnic_profile => {:id => profile_id},
+  #     :mac_address  => network_hash[:mac_address],
+  #   }.delete_blanks
+  #
+  #   if vnic.nil?
+  #     nics_service.add(OvirtSDK4::Nic.new(options))
+  #   else
+  #     nics_service.nic_service(vnic[:id]).update(options)
+  #   end
+  # end
```
borod108: please remove the commented code.

spec/models/manageiq/providers/redhat/infra_manager/provision/configuration/network_spec.rb
```diff
@@ -24,95 +23,227 @@
...
+          set_vnic_profile
 
-    it "should create a new adapter with an optional MAC address" do
-      @task.options[:networks] = [{:network => network_name, :mac_address => mac_address}]
+          expect(nic1_service).to receive(:update)
+          expect(nic2_service).to receive(:update)
+
+          @task.configure_network_adapters
+
+          expect(@task.options[:networks]).to eq([
```
borod108: I think it should be to match_array(...)

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/network.rb
AlonaKaplan: done.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisionInfraWorkflow
   include CloudInitTemplateMixin
 
+  VERSION_4_HASH = {:version => 4}.freeze
```
AlonaKaplan: done.

spec/models/manageiq/providers/redhat/infra_manager/provision/configuration/network_spec.rb
AlonaKaplan: done

masayag:
@AlonaKaplan pls fix typo in message:

 

> In case of **overdding** template vnics. All template vnics will be removed

-->  overriding

masayag:
@AlonaKaplan according to https://github.com/ManageIQ/manageiq/pull/14991/files#diff-082fa3246dc591feb51f4ff99c4080c3R421 the commit message is outdated: 

> Changing the 'vLan' label to 'Virtual Nic Profile'

Seems that the new title of that dialog is **Network**

masayag:
@miq_bot add_label enhancement

PR #31, opened 2017-05-08
Fix host targeted refresh

masayag:
The PR handles two issues with host targeting refresh:
1. When targeting a refresh of a single host, no vms or templates are
reported. Therefore the vm or template collected array should be handled
in case of nil.

2. In V3 the host resources retrieved by
/ovirt-engine/api/hosts/{host:id} contains the statistics element. The
statistics element contains a reference to the host element. The xpath
used to collect similar host entries considered the host sub-element
as a resource to collect, which lead to collecting the same host
multiple times. By specifying the exact top level host path, this will
be avoided.

https://bugzilla.redhat.com/show_bug.cgi?id=1448690

### DIFF
--- a/lib/manageiq/providers/ovirt/legacy/inventory.rb
+++ b/lib/manageiq/providers/ovirt/legacy/inventory.rb
@@ -92,8 +92,8 @@ module ManageIQ
             @service.get_resource_by_ems_ref(uri_suffix, element_name)
           end
 
-          def get_resources_by_uri_path(uri_suffix, element_name = nil)
-            @service.get_resources_by_uri_path(uri_suffix, element_name)
+          def get_resources_by_uri_path(uri_suffix, element_name = nil, xpath = nil)
+            @service.get_resources_by_uri_path(uri_suffix, element_name, xpath)
           end
 
           def refresh
@@ -166,13 +166,14 @@ module ManageIQ
 
           def collect_primary_targeted_jobs(jobs)
             results = collect_in_parallel(jobs) do |key, ems_ref|
+              xpath = key == :host ? '/host' : nil
               if ems_ref.kind_of?(Array)
-                ems_ref.flat_map { |item| get_resources_by_uri_path(item) rescue Array.new }
+                ems_ref.flat_map { |item| get_resources_by_uri_path(item, nil, xpath) rescue Array.new }
               elsif ems_ref.kind_of?(Hash)
                 collection, element_name = ems_ref.first
                 standard_collection(collection, element_name, true)
               else
-                get_resources_by_uri_path(ems_ref) rescue Array.new
+                get_resources_by_uri_path(ems_ref, nil, xpath) rescue Array.new
               end
             end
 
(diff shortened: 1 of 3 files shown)


masayag:
@jhernand @borod108  please review

masayag:
This patch depends on https://github.com/ManageIQ/ovirt/pull/82

masayag:
[Bug 1448690 - RHV provider refresh fail on "undefined method `split' for nil:NilClass"](https://trello.com/c/Lfqv3Y3j/143-bug-1448690-rhv-provider-refresh-fail-on-undefined-method-split-for-nil-nilclass)

lib/manageiq/providers/ovirt/legacy/inventory.rb
```diff
@@ -166,13 +166,14 @@ def collect_primary_jobs(jobs)
 
           def collect_primary_targeted_jobs(jobs)
             results = collect_in_parallel(jobs) do |key, ems_ref|
+              xpath = '/host' if key == :host
```
jhernand: If `key` is not `:host` then `xpath` won't be defined. Won't the below code fail then because the local variable doesn't exist?

masayag:
@miq-bot add_label bug

masayag:
@miq-bot add_label fine/yes

lib/manageiq/providers/ovirt/legacy/inventory.rb
masayag: xpath will be defined with 'nil' as its value.
i.e.
x = 'aa' if false
returns x with nil value.

jhernand: I am an ignorant :-( .

Looks good to me.

bdunne: Normally we would prefer something like this to make it more obvious that the `nil` is intentional.
```ruby
xpath = key == :host ? '/host' : nil
```

masayag: @bdunne I'll change it to be more clear.

lib/manageiq/providers/ovirt/legacy/inventory.rb
```diff
@@ -92,8 +92,8 @@ def get_resource_by_ems_ref(uri_suffix, element_name = nil)
             @service.get_resource_by_ems_ref(uri_suffix, element_name)
           end
 
-          def get_resources_by_uri_path(uri_suffix, element_name = nil)
-            @service.get_resources_by_uri_path(uri_suffix, element_name)
+          def get_resources_by_uri_path(uri_suffix, element_name = nil, xpath = nil)
```
bdunne: I released `ovirt v0.17.0`, please include that change in the gemspec.

masayag: Added.

chrispy1:
@miq-bot add_label blocker

simaishi:
@masayag @oourfali  this is a bit tricky one to backport.. this needs to be split to a commit to `manageiq` repo and another commit to `manageiq-gems-pending` repo.

I will cherry-pick this PR and make a commit to  `manageiq` repo for `app/models/manageiq/providers/redhat/infra_manager/refresh/parse/parser.rb`.

The other 2 files need to go to `manageiq-gems-pending`, but it's not clean a cherry-pick.  The files are renamed and also the content is a bit different which flags as conflict... Can you please create a Fine PR for those files?

cc @bdunne

masayag:
@simaishi I'll create a PR shortly

simaishi:
Fine backport details for `parser.rb`:

```
$ git log -1
commit 5dbf87a82402dd39449aa10778d170e4d2e6982c
Date:   Mon May 8 21:15:52 2017 +0300

    Merge pull request #31 from masayag/fix_host_targeted_refresh
    
    Fix host targeted refresh
    (cherry picked from commit b375913fa371772dc505fad9c8761d189fedecc3)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1449193
```

simaishi:
The other 2 files backported to Fine via https://github.com/ManageIQ/manageiq-gems-pending/issues/155

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/16633

PR #32, opened 2017-05-08
Remove Job#guid from logging proxy info.

yrudman:
There is no need to include `job.guid` into error message when proxy for smart state analysis not available - proxy is instance of `MiqServer` and does not depend on job instance.

This PR is follow-up for https://github.com/ManageIQ/manageiq/pull/15009

@miq-bot add-label technical debt

\cc @jrafanie


oourfali:
@miq-bot assign @masayag

PR #35, opened 2017-05-12
Graph refresh - target vm

pkliczewski:
Collector and parser implementations

RFE link: https://bugzilla.redhat.com/1466172

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -18,6 +18,8 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
   supports :provisioning
   supports :refresh_new_target
 
+  has_many :ems_custom_attributes, :through => :vms_and_templates
+
   def refresher
     Refresh::RefresherBuilder.new(self).build
   end
@@ -76,7 +78,7 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
 
   def vm_reconfigure(vm, options = {})
     ovirt_services_for_reconfigure = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Builder.new(self)
-      .build(:use_highest_supported_version => true).new(:ems => self)
+                                                                                                      .build(:use_highest_supported_version => true).new(:ems => self)
     ovirt_services_for_reconfigure.vm_reconfigure(vm, options)
   end
 
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -18,8 +18,6 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
   supports :provisioning
   supports :refresh_new_target
 
-  has_many :ems_custom_attributes, :through => :vms_and_templates
-
   def refresher
     Refresh::RefresherBuilder.new(self).build
   end
@@ -78,7 +76,7 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
 
   def vm_reconfigure(vm, options = {})
     ovirt_services_for_reconfigure = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Builder.new(self)
-                                                                                                      .build(:use_highest_supported_version => true).new(:ems => self)
+      .build(:use_highest_supported_version => true).new(:ems => self)
     ovirt_services_for_reconfigure.vm_reconfigure(vm, options)
   end
 
--- a/app/models/manageiq/providers/redhat/inventory/collector.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector.rb
@@ -2,7 +2,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   require_nested :InfraManager
   require_nested :TargetCollection
 
-  attr_reader :clusters
+  attr_reader :emsclusters
   attr_reader :networks
   attr_reader :storagedomains
   attr_reader :datacenters
@@ -19,7 +19,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   end
 
   def initialize_inventory_sources
-    @clusters       = []
+    @emsclusters    = []
     @networks       = []
     @storagedomains = []
     @datacenters    = []
@@ -33,7 +33,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
     ::ManageIQ::Providers::Redhat::Inventory::HashCollection
   end
 
-  def collect_clusters
+  def collect_emsclusters
     manager.with_provider_connection(VERSION_HASH) do |connection|
       connection.system_service.clusters_service.list
     end
--- a/app/models/manageiq/providers/redhat/inventory/collector.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector.rb
@@ -2,7 +2,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   require_nested :InfraManager
   require_nested :TargetCollection
 
-  attr_reader :emsclusters
+  attr_reader :ems_clusters
   attr_reader :networks
   attr_reader :storagedomains
   attr_reader :datacenters
@@ -19,7 +19,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   end
 
   def initialize_inventory_sources
-    @emsclusters    = []
+    @ems_clusters   = []
     @networks       = []
     @storagedomains = []
     @datacenters    = []
@@ -33,7 +33,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
     ::ManageIQ::Providers::Redhat::Inventory::HashCollection
   end
 
-  def collect_emsclusters
+  def collect_ems_clusters
     manager.with_provider_connection(VERSION_HASH) do |connection|
       connection.system_service.clusters_service.list
     end
--- a/app/models/manageiq/providers/redhat/inventory/collector.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector.rb
@@ -3,7 +3,6 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   require_nested :TargetCollection
 
   attr_reader :clusters
-  attr_reader :vmpools
   attr_reader :networks
   attr_reader :storagedomains
   attr_reader :datacenters
@@ -11,7 +10,9 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
   attr_reader :vms
   attr_reader :templates
 
-  def initialize(_manager, _target)
+  VERSION_HASH = {:version => 4}.freeze
+
+  def initialize(manager, _target)
     super
 
     initialize_inventory_sources
@@ -19,7 +20,6 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManagerRefresh::Invent
 
   def initialize_inventory_sources
     @clusters       = []
-    @vmpools        = []
     @networks       = []
     @storagedomains = []
     @datacenters    = []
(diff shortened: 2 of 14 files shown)


app/models/manageiq/providers/redhat/inventory/collector.rb
```diff
@@ -0,0 +1,102 @@
...
+    @vms            = []
+    @templates      = []
+  end
+
+  def hash_collection
+    # TODO: check whether needed
+    ::ManageIQ::Providers::Redhat::Inventory::HashCollection
+  end
+
+  def collect_clusters
```
Ladas: nit: we could even drop the collect_

might be that collector.clusters.each reads better than collector.collect_clusters.each

app/models/manageiq/providers/redhat/inventory/collector/target_collection.rb
```diff
@@ -0,0 +1,164 @@
...
+    super
+    parse_targets!
+    infer_related_ems_refs!
+
+    # Reset the target cache, so we can access new targets inside
+    target.manager_refs_by_association_reset
+  end
+
+  def clusters
+    if target.kind_of? VmOrTemplate
```
Ladas: right, seems like you still count with only 1 target being here.

If should be that the target is a TargetCollection, then target.targets could contain VmOrTemplate object

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+    storagedomains
+    hosts
+    vms
+
+    $rhevm_log.info("#{log_header}...Complete")
+  end
+
+  def clusters
+    collector.clusters.each do |cluster|
+      persister.resource_pools.find_or_build(r_id).assign_attributes(
```
Ladas: the r_id does not seem to be defined?

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+        :uid_ems    => "#{cluster.id}_respool",
+        :is_default => true,
+      )
+
+      ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(cluster.href)
+
+      persister.clusters.find_or_build(cluster.id).assign_attributes(
+        :ems_ref       => ems_ref,
+        :ems_ref_obj   => ems_ref,
+        :uid_ems       => cluster.id,
```
Ladas: in the persistor, you have the default unique uuid (manager_ref attribute) as ems_ref, which is default. The unique uuid attribute should be what you use in persister.clusters.find_or_build(unique_uuid/manager_ref)

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+      storage_type = storagedomain.dig(:storage, :type).upcase
+      location = if storage_type == 'NFS' || storage_type == 'GLUSTERFS'
+                   "#{storagedomain.dig(:storage, :address)}:#{storagedomain.dig(:storage, :path)}"
+                 else
+                   logical_units = storagedomain.dig(:storage, :volume_group, :logical_units)
+                   logical_unit =  logical_units && logical_units.first
+                   logical_unit && logical_unit.id
+                 end
+
+      free        = storagedomain.try(:available).to_i
```
Ladas: you can probably put these  one liners directly to the hash below, to spare some LOC

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+
+      ipmi_address = nil
+      if host.dig(:power_management, :type).to_s.include?('ipmi')
+        ipmi_address = host.dig(:power_management, :address)
+      end
+
+      host_os_version = host.dig(:os, :version)
+      ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(host_inv.href)
+
+      clusters = persister.clusters.lazy_find(host.dig(:cluster, :id))
```
Ladas: if you want to access the clusters directly do:

cluster = persister.clusters.find(host.dig(:cluster, :id)) # should return just 1

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+      dc_ids = clusters.collect { |c| c[:datacenter_id] }.uniq
+      storage_ids = []
+      dcs.each do |dc|
+        if dc_ids.include? dc.id
+          storage_ids << collector.collect_dc_domains(dc).to_miq_a.collect { |s| s[:id] }
```
Ladas: would be nice to have a TODO if we could solve this lazily, maybe we can't

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+        :connection_state => connection_state,
+        :power_state      => power_state,
+        :ems_cluster      => clusters,
+        :ipmi_address     => ipmi_address,
+        :storages         => storage_ids
+      )
+
+      host_operating_systems(persister_host, host, hostname)
+      networks = collector.collect_networks
+      switchs(persister_host, host, nics, networks)
```
Ladas: nit: plural of switch should be switches

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+        :subnet_mask => ip.netmask,
+      )
+    end
+  end
+
+  def host_guest_devices(persister_hardware, nics, networks)
+    nics.to_miq_a.each do |nic|
+      network = network_from_nic(nic, host, networks)
+
+      if network
```
Ladas: switch_uid = network.try(:id) || network_name

should result in the same thing

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+      )
+
+      lans(persister_switch, network)
+    end
+  end
+
+  def network_for_nic(nic, host, networks)
+    network_id = nic.dig(:network, :id)
+    if network_id
+      network = networks.detect { |n| n.id == network_id }
```
Ladas: if you cycle this through a lot of hosts&nics, might be better to avoid .detect which is O(n)

so somewhere above you could do indexed_networks = networks.index_by(:id)
and here just do indexed_networks[:network_id], which is O(1)

although I don't see the network_for_nic called anywhere?

Ladas: btw. this would be in general, I've seen that Ovirt has ineffective parser, which is most likely caused by these O(n) lookups, which result easily in O(n^2) when called in another cycle.

So would be nice to also fix parser performance for the graph refresh. :-) I've seen that parsing took the same time as saving to the DB for the old refresh, it should take like 1% of the time we need to save to the DB

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+    else
+      network_name = nic.dig(:network, :name)
+      cluster_id = host.dig(:cluster, :id)
+      cluster = persister.clusters.lazy_find(cluster_id)
+      datacenter_id = cluster.dig(:data_center, :id)
+      network = networks.detect { |n| n.name == network_name && n.dig(:data_center, :id) == datacenter_id }
```
Ladas: a crusade to kill all .detect :-)

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+      operating_systems(persister_vm, vm)
+      custom_attributes(persister_vm, vm)
+    end
+  end
+
+  def storages(vm)
+    storages = []
+    collector.attached_disks(vm).to_miq_a.each do |disk|
+      disk.storage_domains.to_miq_a.each do |sd|
+        storages << persister.storages.lazy_find(:ems_ref => sd.id)
```
Ladas: lazy find knows only strings now and can take only the unique uuid (if needed we could extend it to teach it multiple indexes)

so persister.storages.lazy_find(sd.id)

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+  end
+
+  def snapshots(persister_vm, vm)
+    snapshots = collector.collect_snapshots(vm)
+    parent_id = nil
+    snapshots.each_with_index do |snapshot, idx|
+      name = description = snapshot.description
+      name = "Active Image" if name[0, 13] == '_ActiveImage_'
+
+      persister.snapshots.find_or_build_by(persister_vm).assign_attributes(
```
Ladas: find_or_build

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+        :current     => idx == snapshots.length - 1,
+      )
+      parent_id = snapshot.id
+    end
+  end
+
+  def operating_systems(persister_vm, vm)
+    guest_os = vm.dig(:os, :type)
+
+    persister.operating_systems.find_or_build_by(persister_vm).assign_attributes(
```
Ladas: find_or_build

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,515 @@
...
+      :system_type  => vm.type
+    )
+  end
+
+  def custom_attributes(persister_vm, vm)
+    custom_attrs = vm.try(:custom_properties)
+    return result if custom_attrs.nil?
+
+    custom_attrs.each do |ca|
+      persister.custom_attributes.find_or_build_by(persister_vm).assign_attributes(
```
Ladas: find_or_build

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
pkliczewski: Will move only those which are used in one place

app/models/manageiq/providers/redhat/inventory/collector/target_collection.rb
```diff
@@ -8,33 +8,42 @@ def initialize(_manager, _target)
     target.manager_refs_by_association_reset
   end
 
-  def clusters
-    if target.kind_of? VmOrTemplate
-      return collect_emsclusters
+  def ems_clusters
+    clusters = []
+    if target.targets.any? { |t| t.kind_of? VmOrTemplate }
```
Ladas: right, so for the saving to work properly, the Collector and Persistor should use the same: references(:ems_clusters)

references(:ems_clusters) => [:cluster_ems_ref_1, :cluster_ems_ref_2]
So what it will then say, is collect me only  [:cluster_ems_ref_1, :cluster_ems_ref_2] from the API. And when saving, compare it to [:cluster_ems_ref_1, :cluster_ems_ref_2] in the MIQ DB.

so Collector and Persistor should be aligned like this

pkliczewski:
@borod108 @masayag @jhernand Please take a look

pkliczewski:
@matobet please take a look

pkliczewski:
The reason for failed test is missing default manager which is provided by dependent PR

pkliczewski:
[Graph refresh](https://trello.com/c/uid9QN8b/161-graph-refresh)

pkliczewski:
@miq-bot assign @oourfali

pkliczewski:

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -76,7 +76,7 @@ def self.provision_class(via)
 
   def vm_reconfigure(vm, options = {})
     ovirt_services_for_reconfigure = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Builder.new(self)
-      .build(:use_highest_supported_version => true).new(:ems => self)
+                                                                                                      .build(:use_highest_supported_version => true).new(:ems => self)
```
borod108: extra line?

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -0,0 +1,519 @@
...
+      tag_value = network.try(:vlan).try(:id)
+    else
+      uid = name = network_name unless network_name.nil?
+    end
+
+    if uid.nil?
+      return
+    end
+
+    persister.lans..find_or_build_by(
```
borod108: ..

PR #39, opened 2017-05-24
Remove dependence on ovirt-gem from credentials verification

borod108:
The credentials verification process used the ovirt gem even when the
"use_ovirt_engine_sdk" setting was turned on.
In addition it was catching an exception that was defined in the ovirt gem.
Both issues are now fixed.


borod108:
@masayag please review.

borod108:
@miq-bot assign @oourfali

PR #47, opened 2017-06-02
v2v: Add support for post-import network configuration

matobet:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1404920

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
@@ -32,8 +32,21 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
     )
   end
 
+  def configure_imported_vm_networks(vm_id)
+    _log.info("Configuring networks for VM: #{vm_id}")
+
+    vm = Vm.find(vm_id)
+    dc = vm.ems_cluster.parent_datacenter
+
+    with_provider_connection :version => 4 do |conn|
+      vm_network_id = first_vm_network_id(conn, dc.uid_ems)
+      vnic_profile_id = vnic_profile_id_by_network_id(conn, vm_network_id)
+      set_all_vm_nics_to_profile_id(conn, vm.uid_ems, vnic_profile_id)
+    end
+  end
+
   def submit_import_vm(userid, source_vm_id, target_params)
-    task_id = queue_import_vm(userid, source_vm_id, target_params)
+    task_id = queue_self_method_call(userid, 'Import VM', 'import_vm', source_vm_id, target_params)
     task = MiqTask.wait_for_taskid(task_id)
 
     task.task_results
@@ -41,6 +54,13 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
 
   def validate_import_vm
     Gem::Version.new(api_version) >= Gem::Version.new('4')
+  end
+
+  def submit_configure_imported_vm_networks(userid, vm_id)
+    task_id = queue_self_method_call(userid, "Configure imported VM's networks", 'configure_imported_vm_networks', vm_id)
+    task = MiqTask.wait_for_taskid(task_id)
+
+    task.task_results
   end
 
   private
@@ -54,9 +74,9 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
     end
   end
 
-  def queue_import_vm(userid, source_vm_id, target_params)
+  def queue_self_method_call(userid, action, method_name, *args)
     task_options = {
-      :action => 'Import VM',
+      :action => action,
       :userid => userid
     }
 
@@ -66,13 +86,13 @@ module ManageIQ::Providers::Redhat::InfraManager::VmImport
       # and cause a deadlock in subsequent waiting on this task from synchronous context
       :zone        => my_zone,
       :class_name  => self.class.name,
-      :method_name => 'import_vm',
+      :method_name => method_name,
       :instance_id => id,
       :role        => 'ems_operations',
-      :args        => [source_vm_id, target_params]
+      :args        => args
     }
 
-    _log.info("Queueing import of VM ID: #{source_vm_id} by user #{userid}")
+    _log.info("Queueing '#{method_name}' with args #{args} by user #{userid}")
     MiqTask.generic_action_with_callback(task_options, queue_options)
   end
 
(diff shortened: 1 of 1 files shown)


matobet:
Another from the series of v2v follow-ups, @borod108 @pkliczewski @jhernand please review

matobet:
@miq-bot assign @jhernand

matobet:
@miq-bot add_label fine/yes

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -116,4 +158,22 @@ def select_host(vm)
   def password_auth(provider)
     provider.authentication_userid_passwords.first
   end
+
+  def get_first_vm_network_id(conn, dc_id)
+    conn.system_service.data_centers_service.data_center_service(dc_id)
+        .networks_service.list.select { |net| net.usages.include? 'vm' }.first.id
```
jhernand: You can use `.detect { ... }` instead of `.select { ... }.first`.

There may be no network that satisfies the condition, so this needs to check if the result is `nil`.

Use [`OvirtSDK::NetworkUsage::VM`](http://www.rubydoc.info/gems/ovirt-engine-sdk/OvirtSDK4/NetworkUsage) instead of `'vm'`, if possible.

The results of the `list` method don't have any guaranteed order, so this may produce a result today and a different result tomorrow. I'd suggest that you first get all the networks, sort them (by name) and then get the first. That way the result will be predictable.

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -116,4 +158,22 @@ def select_host(vm)
...
+  end
+
+  def get_vnic_profile_id_by_network_id(conn, network_id)
+    conn.system_service.networks_service.network_service(network_id).vnic_profiles_service.list.first.id
```
jhernand: Like in the previous comment, I'd suggest to sort the profiles by name, to make the result predictable.

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -116,4 +158,22 @@ def select_host(vm)
...
+  end
+
+  def set_all_vm_nics_to_profile_id(conn, vm_id, profile_id)
+    vm_nics_service = conn.system_service.vms_service.vm_service(vm_id).nics_service
+    vm_nics_service.list.map(&:id).each do |nic_id|
+      vm_nics_service.nic_service(nic_id).update(
+        OvirtSDK4::Nic.new(:vnic_profile => OvirtSDK4::VnicProfile.new(:id => profile_id))
```
jhernand: This can be shortened, if you want:

```ruby
...update(
  vnic_profile: {
    id: profile_id
  }
)
```

The SDK translates that hash into the same object that you are explicitly creating. Up to you, both approaches are perfectly OK.

I also find the use of `.map(&:id)` a bit harder to read. Wouldn't it be easier to do this?

```ruby
vm_nics_service.list.each do |nic|
  vm_nics_service.nic_service(nic.id).update(
    ...
  )
end

File: app/models/manageiq/providers/redhat/infra_manager/vm_import.rb (hunk 2 above)
Comment (matobet) [inline app/models/manageiq/providers/redhat/infra_manager/vm_import.rb]: done

File: app/models/manageiq/providers/redhat/infra_manager/vm_import.rb (hunk 1 above)
Comment (matobet) [inline app/models/manageiq/providers/redhat/infra_manager/vm_import.rb]: done, optimized with suggested sorting to `.min_by(&:name)`

**Comment by matobet** [comment]
removed some code duplication from the task queue invocation code

File: app/models/manageiq/providers/redhat/infra_manager/vm_import.rb [hunk 4]
```diff
@@ -43,6 +56,13 @@ def validate_import_vm
     Gem::Version.new(api_version) >= Gem::Version.new('4')
   end
 
+  def submit_configure_imported_vm_networks(userid, vm_id)
```
gmcculloug: Continuing conversation from https://github.com/ManageIQ/manageiq-automation_engine/pull/35#discussion_r120493827.

The caller must know the VM so I feel like these configure methods belong on the VM and not the provider.  I know this is intended for imported VMs but would assume that it would work on any VM so the naming again seems off.

What I would like to understand is if this method can be moved to the appropriate VM model (called without `vm_id`) and renamed to something like `configure_networks`?  Then the method it calls would also be moved and renamed to something better.

matobet: @gmcculloug actually this piece of code is "fixing" the state of the post-imported VM that it contains NICs with no network assigned. Logic here is best effort and tries to pick appropriate network to assign to all of them, to make the VM at least somewhat useful after the conversion process. Calling this logic upon a "normal" (properly configured, existing) VM could only be detrimental to its mode of operation. So I'm not really sure about moving it to the VM model..

@borod108 any opinions

borod108: I tend to agree with @matobet this logic should not normally be called on a VM, and the only reason it might be called is as part of this process.
Of course its more procedural than OOP but it seems to me it will be easier to understand this way.

durandom: TBH this method should neither be on a Vm nor on the Manager but on a VmImporter class. But we havent embraced something like ServiceObjects - https://github.com/ManageIQ/manageiq/issues/12902
So, you'll find both in our codebase. 🤷‍♂️ I'd say keep it here, so its at least in a `VmImport` module...

But `userid` should be `user_id`, to be consistent.

matobet: @durandom I believe it was pointed to me there is a strict distinction between `user.id` (id of the `User` ActiveRecord) and `userid` in terms of the "login credential" of the user to the external system. This is to the best of my understanding the latter.

gmcculloug: @matobet You are correct.  You are passing the `userid` column from the `User` class.  (Not to be confused with the `id` column from the user class.)  😕 

See usage here: https://github.com/ManageIQ/manageiq-content/pull/123/files#diff-20380aa1ea403e71a951c411b93702f6R22

As far as naming goes, can we drop the `submit_` prefix?

matobet: @gmcculloug than we have to come up with different name for `configure_imported_vm_networks` https://github.com/ManageIQ/manageiq-providers-ovirt/pull/47/files#diff-9093cd3391ba88c199808e385f9304e0R35 since the `submit_` method actually only enqueues the actual `configure_imported_vm_networks` method.

matobet: This `do_stuff` and `submit_do_stuff` is actually used for import vm `submit_import_vm` + `import_vm` and I believe was originally conceived together with @mkanoor in the initial v2v design phase. Here I'm just trying to maintain consistency

app/models/manageiq/providers/redhat/infra_manager/vm_import.rb
```diff
@@ -119,4 +139,20 @@ def select_host(vm)
   def password_auth(provider)
     provider.authentication_userid_passwords.first
   end
+
+  def get_first_vm_network_id(conn, dc_id)
```
gmcculloug: I thought rubocop use to complain about using get/set method names.  In the past we tried to avoid them.  Ref https://github.com/bbatsov/ruby-style-guide#accessor_mutator_method_names.

These two `get_` methods look straightforward to change.  The `set_` here I could live with.

Glad to see these methods already in the private section of the class.  👍

matobet: @gmcculloug done

simaishi:
Fine backport details:

```
$ git log -1
commit 89a54856f076351a196bca6fb3b2a1b916bf890b
Date:   Tue Jun 13 11:31:40 2017 +0300

    Merge pull request #47 from matobet/v2v-networks
    
    v2v: Add support for post-import network configuration
    (cherry picked from commit 43d34e4399d05c4218d8c48ee672ccbfe4e2d71a)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459996
```

PR #50, opened 2017-06-22
Fix provisioning from PXE with the OvirtSDK

borod108:
Part of the PXE provisioning process which had to do with finding the mac address
was broken.

https://bugzilla.redhat.com/show_bug.cgi?id=1454861

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
@@ -227,9 +227,24 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       end
     end
 
-    def find_mac_address_on_network(nics, network, log)
+    def get_mac_address_of_nic_on_requested_vlan(args)
+      network = find_network_in_cluster(args[:value_of_vlan_option], args[:dest_cluster])
+      return nil if network.nil?
+
+      find_mac_address_on_network(network, args[:destination])
+    end
+
+    def find_network_in_cluster(network_name, dest_cluster)
+      network = cluster_find_network_by_name(dest_cluster.ems_ref, network_name)
+
+      _log.warn "Cannot find network name=#{network_name}" if network.nil?
+      network
+    end
+
+    def find_mac_address_on_network(network, destination)
+      nics = nics_for_vm(destination)
       nic = nics.detect { |n| n.network.try(:id) == network[:id] }
-      log.warn "Cannot find NIC with network id=#{network[:id].inspect}" if nic.nil?
+      _log.warn "Cannot find NIC with network id=#{network[:id].inspect}" if nic.nil?
       nic && nic[:mac] && nic[:mac][:address]
     end
 
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -347,14 +347,17 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       URI(ems_ref).path.split('/').last
     end
 
-    def find_mac_address_on_network(nics, network, log)
-      ext_management_system.with_provider_connection(VERSION_HASH) do |connection|
-        nic = nics.detect do |n|
-          connection.follow_link(n.vnic_profile).network.id == network[:id]
-        end
-        log.warn "Cannot find NIC with network id=#{network[:id].inspect}" if nic.nil?
-        nic && nic.mac && nic.mac.address
+    def get_mac_address_of_nic_on_requested_vlan(args)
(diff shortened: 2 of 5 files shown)


borod108:
@miq-bot add_label fine/no
@masayag @pkliczewski @jhernand any of you can review?

borod108:
@miq-bot add_label wip

borod108:
@masayag can you please review?

borod108:
@miq-bot remove_label wip

validated for both v3 and v4.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -227,9 +227,24 @@ def method_missing(method_name, *args, &block)
       end
     end
 
-    def find_mac_address_on_network(nics, network, log)
+    def get_mac_address_of_nic_on_requested_vlan(args)
```
AlonaKaplan: I would pass explicit arguments and not args. It is more readable.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -227,9 +227,24 @@ def method_missing(method_name, *args, &block)
...
+      network = find_network_in_cluster(args[:value_of_vlan_option], args[:dest_cluster], args[:source_template])
+      return nil if network.nil?
+
+      find_mac_address_on_network(network, args[:destination])
+    end
+
+    def find_network_in_cluster(network_name, dest_cluster, source)
+      network = source.ext_management_system.ovirt_services.cluster_find_network_by_name(dest_cluster.ems_ref, network_name)
```
AlonaKaplan: The method ('cluster_find_network_by_name') is in this class, why don't you call it directly?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
borod108: There are advantages I agree but this allows to path arguments in any order and change them later without breaking the interface. Its a common practice.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
borod108: Because thia instance is initialized with the ems from the destination vm and that one is run on the ema from the source template. When running the code i saw the ems on the vm is not set yet (probably since the vm is only beimg provisioned) and it is required for this method to work so this actually initialzes an instance with ems of the source template and calls the method on it.

AlonaKaplan: 1. Can't you solve it by invoking the code (in network.rb) using the template's ems? (If the source ems is not initialized yet, it seems to be the correct approach).
2. How 'nics_for_vm(..)' which is called directly from 'find_mac_address_on_network(..)' is working?
Doesn't it need the ems as well to get the connection?

borod108:
[Fix PXE with Alonas Changes](https://trello.com/c/1D1ReK5z/162-fix-pxe-with-alonas-changes)

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
borod108: 1. I think its proper to put this method here since in v4 it is not used so it does not make sense to have it in network.rb (actually this is exactly the reason why we split both of them - in v3 we find network by name, in v4 we do not need it since we have the vnic_profile, you were the one who explained that to me...)
2. nics_for_vm does not use the ext_management_system, it make a call on the vm directly.  The cluster_find_network_by_name is used in other places, so I would rather not replacement it to use vm.

AlonaKaplan: 1. I didn't mean you need to move the method to another file. In network.rb invoke **source**.ext_management_system.ovirt_services.get_mac_address_of_nic_on_requested_vlan(args) instead of destination.ext_management_system.ovirt_services.get_mac_address_of_nic_on_requested_vlan(args).
2. 'nics_for_vm(..)' invokes
'vm.with_provider_object' which invokes
'connection_source(..)' which invokes
def connection_manager
    try(:ext_management_system) || try(:manager)
 end
So, IIUC, 'ext_management_system' of the vm is invoked.
So why here it is working and in your case it is not?

borod108: @AlonaKaplan yes, you are right :)
This was probably true for a diffrent phase of the configuration, and stack in my mind. I changed the code according to your suggestion and it does look better :)

borod108:
@miq-bot assign @oourfali can you please merge?

borod108:
@miq-bot assign @oourfali
can you please merge?

PR #61, opened 2017-07-16
Support memory limit

masayag:
Memory limit feature was introduced in RHV 4.1.
It is already supported by manageiq as part of the vm provision process.
The patch adds support for total memory for providers of version 4.1 and
above, both for VM provision and for the VM refresh process.

https://bugzilla.redhat.com/show_bug.cgi?id=1461560

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -146,4 +146,9 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
   def supports_migrate_for_all?(vms)
     vms.map(&:ems_cluster).uniq.compact.size == 1
   end
+
+  def version_higher_than?(version)
+    ems_version = api_version[/\d+\.\d+/x]
+    Gem::Version.new(ems_version) >= Gem::Version.new(version)
+  end
 end
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/configuration/container.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/configuration/container.rb
@@ -9,9 +9,12 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration::Cont
 
   def configure_memory(rhevm_vm)
     vm_memory = get_option(:vm_memory).to_i * 1.megabyte
-    return if vm_memory.zero?
-    _log.info "Setting memory to:<#{vm_memory.inspect}>"
-    rhevm_vm.update_memory!(vm_memory)
+    memory_limit = get_option(:memory_limit).to_i * 1.megabyte
+    return if vm_memory.zero? && memory_limit.zero?
+
+    limit_message = ", total: <#{memory_limit}>" unless memory_limit.zero?
+    _log.info("Setting memory to:<#{vm_memory.inspect}>#{limit_message}")
+    rhevm_vm.update_memory!(vm_memory, memory_limit)
   end
 
   def configure_memory_reserve(rhevm_vm)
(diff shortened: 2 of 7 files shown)


app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -108,4 +108,19 @@ def filter_allowed_hosts(all_hosts)
     ovirt_services = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Builder.new(ems).build(:use_highest_supported_version => true).new(:ems => ems)
     ovirt_services.filter_allowed_hosts(self, all_hosts)
   end
+
+  def validate_memory_limit(_field, values, dlg, fld, _value)
```
masayag: usually, generic validation methods are defined in a higher parent-class, however, since this method contains a specific RHV logic, I defined it in this file.

masayag:
@miq-bot assign @borod108

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -108,4 +108,19 @@ def filter_allowed_hosts(all_hosts)
...
+    limited = get_value(values[:memory_limit])
+    return nil unless limited
+
+    ems_version = source_ems.api_version[/\d+\.\d+/x]
```
borod108: I think this and the next line should be in a method, like "supported_in_version?" or something like that.

app/models/manageiq/providers/redhat/infra_manager/provision/configuration/container.rb
```diff
@@ -9,9 +9,12 @@ def configure_container_description(rhevm_vm)
...
   def configure_memory(rhevm_vm)
     vm_memory = get_option(:vm_memory).to_i * 1.megabyte
-    return if vm_memory.zero?
-    _log.info "Setting memory to:<#{vm_memory.inspect}>"
-    rhevm_vm.update_memory!(vm_memory)
+    memory_limit = get_option(:memory_limit).to_i * 1.megabyte
+    return if vm_memory.zero? && memory_limit.zero?
+
+    limit_message = ", total: <#{memory_limit}>" unless memory_limit.zero?
+    _log.info "Setting memory to:<#{vm_memory.inspect}>#{limit_message}"
```
borod108: rubocop wanted parentheses.

masayag: done.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
masayag: done

masayag:
[Bug 1461560 - Provisioning to RHV 4.1 Max Memory Size Needs to be Adjusted as Necesary](https://trello.com/c/5L3fZRQA/160-bug-1461560-provisioning-to-rhv-41-max-memory-size-needs-to-be-adjusted-as-necesary)

masayag:
@borod108 while this patch is verified to work with use_ovirt_engine_sdk = true, the total memory will be ignored if using v3 (use_ovirt_engine_sdk = false).

Do you think we should block this action in such case ? Or just ignore the provided parameter ?

borod108:
@masayag I think it should be either blocked, or at least a warning should be shown or it should be mentioned in the description of this field. But the best option, I think, is to block it...

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -146,4 +146,9 @@ def unsupported_migration_options
   def supports_migrate_for_all?(vms)
     vms.map(&:ems_cluster).uniq.compact.size == 1
   end
+
+  def version_higher_than?(version)
```
masayag: @matobet perhaps this can be useful for 'import vm' version checks

matobet: Exactly! CC @smelamud

masayag:
@borod108 so i added the following to this patch:

1. Nullify memory limit if not using the SDK - this should maintain the UI behavior for RHV >= 4.1 that doesn't use the sdk
2. Throw an error message in case of unsupported RHV version tries to pass a parameter for memory limit

masayag:
@miq-bot assign @oourfali

masayag:
@miq-bot add_label fine/yes

PR #62, opened 2017-07-17
disconnect_storage should be called once

pkliczewski:
During vm remove we used to call disconnect_storage. The first time
it was called as a first step of removal flow and during refresh.
This caused timing issue which caused lack of update to vm.ems_id.
As a result vm state was "unknown".

We need to make sure that disconnect_storage is called only once.

Bug-Url:
https://bugzilla.redhat.com/1468201

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm.rb
@@ -48,25 +48,25 @@ class ManageIQ::Providers::Redhat::InfraManager::Vm < ManageIQ::Providers::Infra
   alias ems_cluster parent_cluster
 
   def disconnect_storage(_s = nil)
-    unless active?
-      return
-    end
-    vm_disks = collect_disks
+    return unless active?
 
-    if vm_disks.blank?
-      storage = nil
-    else
-      vm_storages = ([storage] + storages).compact.uniq
-      storage = vm_storages.select { |store| !vm_disks.include?(store.ems_ref) }
-    end
+    vm_storages = ([storage] + storages).compact.uniq
+    return if vm_storages.empty?
+
+    vm_disks = collect_disks
+    storage = vm_disks.blank? ? nil : vm_storages.select { |store| !vm_disks.include?(store.ems_ref) }
 
     super(storage)
   end
 
   def collect_disks
     return [] if hardware.nil?
-    disks = hardware.disks.map { |disk| "#{disk.storage.ems_ref}/disks/#{disk.filename}" }
-    ext_management_system.ovirt_services.collect_disks_by_hrefs(disks)
+    disks = hardware.disks.map do |disk|
+      unless disk.storage.nil?
+        "#{disk.storage.ems_ref}/disks/#{disk.filename}"
+      end
+    end
+    ext_management_system.ovirt_services.collect_disks_by_hrefs(disks.compact)
   end
 
   def disconnect_inv


pkliczewski:

pkliczewski:
[Bug 1468201 - A deleted VM state do not change to Archived state](https://trello.com/c/dHeu2r8v/177-bug-1468201-a-deleted-vm-state-do-not-change-to-archived-state)

pkliczewski:
@miq-bot assign @borod108

pkliczewski:
@miq-bot add_label fine/yes

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -48,17 +48,18 @@ def parent_cluster
...
+
     vm_disks = collect_disks
 
-    if vm_disks.blank?
-      storage = nil
-    else
-      vm_storages = ([storage] + storages).compact.uniq
-      storage = vm_storages.select { |store| !vm_disks.include?(store.ems_ref) }
-    end
+    storage = if vm_disks.blank?
```
borod108: I think this will work and looks more compact:
`storage = vm_storages.select { |store| !vm_disks.include?(store.ems_ref) } if vm_disks.blank?`

pkliczewski: look reasonable, will update

masayag: I don't think this behave the same as the code that @pkliczewski wrote.
In @pkliczewski there will be an assignment to `storage` in any case.
In your suggestion, the `storage` might have a previous value that won't be erased by your suggested fix.

pkliczewski: @masayag is right after testing the change it doesn't have the same behavior.

pkliczewski: @borod108 sure, we can use ternary expression

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -48,17 +48,13 @@ def parent_cluster
...
-    if vm_disks.blank?
-      storage = nil
-    else
-      vm_storages = ([storage] + storages).compact.uniq
-      storage = vm_storages.select { |store| !vm_disks.include?(store.ems_ref) }
-    end
+    vm_storages = ([storage] + storages).compact.uniq
+    return if vm_storages.empty?
+
+    vm_disks = collect_disks
```
pkliczewski: @masayag No, if the disk is removed it will be not listed. According to my understanding of the code disk.storage always will have a value.

pkliczewski:
@borod108 @masayag Are you OK with this change?

oourfali:
Any other review is needed? Shall we merge?

pkliczewski:
@oourfali I think that we have all the acks we need.

simaishi:

```
$ git log -1
commit 1a0907bce19aed801f3b01bb0c06546b2c8ac134
Date:   Wed Jul 26 11:44:03 2017 +0300

    Merge pull request #62 from pkliczewski/master
    
    disconnect_storage should be called once
    (cherry picked from commit c1257ddd0e200d76a316de03261bf92c2a8cfbfa)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1479481
```

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/16218

PR #65, opened 2017-07-19
Host maintenance

pkliczewski:
There were few maintenance event missing:

USER_VDS_MAINTENANCE_WITHOUT_REASON(620)
VDS_MAINTENANCE_MANUAL_HA(10452)
USER_VDS_MAINTENANCE_MANUAL_HA(10453)

which was unknown by the provider. As a result we do not trigger
refresh and the host status is not updated.


pkliczewski:
@miq-bot assign @borod108

pkliczewski:

pkliczewski:
@oourfali do we want to fix it for v3 as well?

oourfali:
@pkliczewski I'd say that we do.
Unless that's too complex.

pkliczewski:
@oourfali the change would be similar but for [1] and we would need to bump gem version.
Will push a PR and let's have a discussion there.

[1] https://github.com/ManageIQ/ovirt/blob/master/lib/ovirt/event.rb#L316

pkliczewski:
Corresponding v3 fix can be found here: https://github.com/ManageIQ/ovirt/pull/85

app/models/manageiq/providers/redhat/infra_manager/event_fetcher.rb
```diff
@@ -313,6 +313,7 @@ def set_event_name!(event)
     602  => "USER_HOST_MAINTENANCE_MIGRATION_FAILED",
     603  => "HOST_SET_NONOPERATIONAL_IFACE_DOWN",
     605  => "HOST_USED_AS_PROXY",
+    620  => "USER_HOST_MAINTENANCE",
```
masayag: while at it, I'd include the other host maintenance event types (added for host-engine's host maintenance):

- VDS_MAINTENANCE_MANUAL_HA (10452)
- USER_VDS_MAINTENANCE_MANUAL_HA (10453)

pkliczewski: Good idea, I haven't seen those events but will add those here. I think that as some point we need to align this file with [1]. I will add a card to our backlog.

[1] https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/AuditLogType.java

masayag:
One thought about it, do we want to include also the "failed to maintenance" events ? The host might be changed to 'non-responsive' as a result of failed attempt. This has low priority and could be considered later-on.

pkliczewski:
@masayag I think that we already have it here [1]. Do you see any other events that are not there?

[1] https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/event_fetcher.rb#L34

masayag:
@pkliczewski Seems that all of the failed events are covered. 

Please update the commit message and PR message to include the additional events.

PR #66, opened 2017-07-20
add raw_connect class method

jntullo:
The UI is moving towards validating credentials on the queue to allow for validating credentials in separate zones utilizing a `raw_connect` class method, rather than needing to create a temporary EMS for validation.

While most providers have a single `raw_connect` class method, the Ovirt provider does not. This PR essentially adds a wrapper around a couple of the other methods found within the `api_integration` module that allows for different versions of raw connect to be called by accepting a version parameter.

It also decrypts that password before passing it to the existing functions, because passwords will be encrypted now that we will be placing them on the queue. This function will also work with unencrypted passwords.

For more information, see: https://github.com/ManageIQ/manageiq-ui-classic/pull/1580
(will be creating a separate PR from that for infra providers, but it provides the motivation / overall context)

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -129,11 +129,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       # using the `disconnect` method. If `version` is 4 the object is created by the oVirt Ruby SDK, and
       # it is closed using the `close` method.
       begin
-        if connection.respond_to?(:disconnect)
-          connection.disconnect
-        elsif connection.respond_to?(:close)
-          connection.close
-        end
+        self.class.disconnect(connection)
       rescue => error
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -273,6 +277,19 @@ def process_api_features_support
       end
     end
 
+    def raw_connect(options)
+      version = options[:version]
```
borod108: How will the client of this method decide which version to pass here?
We use this https://github.com/jntullo/manageiq-providers-ovirt/blob/3930da25d85fd020cbcc29ecb5a72b142ee933c3/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb#L19
in the connect method.

borod108:
LGTM :) can you please provide a link to the code that will use this method?

jntullo:
@borod108 it will soon be used here: https://github.com/ManageIQ/manageiq/pull/15606 :)

PR #67, opened 2017-07-20
remove invalid rescue statement

jntullo:
When investigating the available methods for https://github.com/ManageIQ/manageiq-providers-ovirt/pull/66, I noticed that the `#verify_credentials_for_rhevm`method was not working and was raising a 

```
NameError:
       uninitialized constant ManageIQ::Providers::Redhat::InfraManager::ApiIntegration::RestClient
```

error because of what appears to be an invalid `rescue`. I looked for the `RestClient` that it may want, but was unable to find it. Perhaps somebody more familiar with Ovirt may be able to point it out, but without this removed currently, the method does not work at all.

Added a test that fails without this change for good measure.

@miq-bot add_label bug

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -5,6 +5,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
   extend ActiveSupport::Concern
 
   require 'ovirtsdk4'
+  require 'ovirt'
 
   included do
     process_api_features_support
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -284,6 +284,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       # values, so we need to check the value passed and make a list only if it won't be empty. If it will be empty then
       # we should just pass 'nil'.
       ca_certs = options[:ca_certs]
+      ca_certs = nil if ca_certs.blank?
       ca_certs = [ca_certs] if ca_certs
 
       url = URI::Generic.build(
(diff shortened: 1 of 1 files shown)


masayag:
@miq-bot assign @borod108

borod108:
@jntullo hi, for some reason it did not reproduce, I tried causing the exception with both v3 and v4. Any clue how I can reproduce this?

Because we do need to catch this exception when using v3 to show a proper error message to the user.

oourfali:
@jntullo can you take a look at the question above?

jntullo:
Hi @borod108 and @oourfali - sorry for the delay.

I was able to reproduce via 
```
[3] pry(main)> ManageIQ::Providers::Redhat::InfraManager.new.verify_credentials_for_rhevm(:version => 'v4', :auth_type => 'foo')
NameError: uninitialized constant ManageIQ::Providers::Redhat::InfraManager::ApiIntegration::RestClient
from /Users/jilliantullo/code/manageiq-providers-ovirt/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb:151:in `rescue in verify_credentials_for_rhevm'
```

and then with this fix, it performs as expected:
```
[1] pry(main)> ManageIQ::Providers::Redhat::InfraManager.new.verify_credentials_for_rhevm(:version => 'v4', :auth_type => 'foo')
PostgreSQLAdapter#log_after_checkout, connection_pool: size: 16, connections: 1, in use: 1, waiting_in_queue: 0
MiqException::MiqEVMLoginError: no credentials defined
```

borod108:
@jntullo Hi, sorry for the delay in response.
1. The spec passes even without the fix.
2. We actually need this exception handling in case we are using the ovirt gem to connect to the provider (in case it is an old provider, version 3.6 for example), the RestClient will be loaded once the ovirt-gem is loaded.
I would propose adding `require 'ovirt'`  to the api_integration.rb file, it seems to solve the problem. What do you think?

jntullo:
closing in favor of https://github.com/ManageIQ/manageiq-providers-ovirt/pull/101

PR #68, opened 2017-07-25
Fix missing variable in disks preloading

borod108:
Add the missing connection variable


borod108:
@pkliczewski please review.

PR #70, opened 2017-07-26
Target template

pkliczewski:
When we create new template we want to target this template during
a refresh.

Bug-Url:
https://bugzilla.redhat.com/1451577

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
@@ -40,6 +40,8 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParsing
         username = ems.ovirt_services.username_by_href(user_href)
       end
 
+      vm_ref = template?(event[:name]) ? ems_ref_from_object_in_event(event[:template]) : ems_ref_from_object_in_event(event[:vm])
+
       # Build the event hash
       {
         :event_type          => event[:name],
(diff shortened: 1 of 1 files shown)


pkliczewski:
@miq-bot assign @borod108

pkliczewski:

pkliczewski:
[Bug 1451577 - [RFE] Targeted refresh or Templates events](https://trello.com/c/FCLCOxwW/181-bug-1451577-rfe-targeted-refresh-or-templates-events)

app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
```diff
@@ -66,7 +68,15 @@ def self.parse_new_target(full_data, message, ems, event_type)
 
       cluster = parse_new_cluster(ems, full_data[:cluster], dc)
       rp      = parse_new_resource_pool(cluster)
-      vm      = parse_new_vm(ems, full_data[:vm], dc, cluster, message, event_type)
+      if event_type.include?("TEMPLATE")
+        vm_data = full_data[:template]
+        cls = 'ManageIQ::Providers::Redhat::InfraManager::Template'
+      else
+        vm_data = full_data[:vm]
+        cls = 'ManageIQ::Providers::Redhat::InfraManager::Vm'
```
borod108: from what I have seen in ruby usually "klass" var name is used.

app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
```diff
@@ -40,6 +40,8 @@ def self.event_to_hash(event_obj, ems_id = nil)
         username = ems.ovirt_services.username_by_href(user_href)
       end
 
+      vm_ref = event[:name].include?("TEMPLATE") ? ems_ref_from_object_in_event(event[:template]) : ems_ref_from_object_in_event(event[:vm])
```
borod108: > event[:name].include?("TEMPLATE")
Is there any way to do this differently? In any case I think this should be moved to a method and maybe when we find a better way to do it refactor it.

app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
pkliczewski: OK

app/models/manageiq/providers/redhat/infra_manager/event_parsing/parser.rb
pkliczewski: There is no other way to do it. I do not think that a method would be any help here.

PR #71, opened 2017-08-02
Fixed hosts filtering when no ems selected

masayag:
When network profiles were added, the default implementation of
filter_allowed_hosts was overridden.
The previous behavior used to return all of the hosts if couldn't
determine the vlan.

Same behavior should be maintained if there isn't ems available in the
context of the current selection.


masayag:
[Fixed hosts filtering](https://trello.com/c/bejOefYQ/186-fixed-hosts-filtering)

masayag:
@miq-bot assign @borod108

oourfali:
Looks good to me

PR #72, opened 2017-08-03
Fix missing number_of_nics on host hardware

borod108:
This field used to be populated in v3 but we missed it in the new parser.
Added it back to v4 parser.


borod108:
@miq-bot assign @masayag

oourfali:
Looks good to me

PR #74, opened 2017-08-06
Make full refresh through OvirSDK use asynchronous calls to oVirt

borod108:
Make full refresh through OvirSDK use asynchronous calls to oVirt


borod108:
@miq-bot assign @masayag 
Please review.

PR #77, opened 2017-08-14
Fix allowed vlans to allow no provider

masayag:
A template might loose its association with a provider when a template
is being removed from the provider. In that case the template will
become archived or orphan and its ext_management_system will return
nil.

The service catalog filters such vms or templates. However, an existing
service doesn't produce a meaningful error in such case.

Letting the service to allow nil value for EMS at dialog load will
produce a descriptive error when trying to provision.

https://bugzilla.redhat.com/show_bug.cgi?id=1471124


masayag:
@miq-bot assign @borod108 
There is another PR to manageiq for fail this action with a proper message.

masayag:
[Bug 1471124 - Delete a Template in RHEV that a Catalog uses, no indication in logs or UI when Catalog Ordered](https://trello.com/c/zLt3JQes/189-bug-1471124-delete-a-template-in-rhev-that-a-catalog-uses-no-indication-in-logs-or-ui-when-catalog-ordered)

PR #79, opened 2017-08-17
Move logging in trap context to a normal signal handler

jrafanie:
Note, this is nearly identical to
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/95

Since ruby 2.0, you cannot use a mutex in a trap context such as writing
a log message.

```
irb(main):001:0> require 'logger'
=> true
irb(main):002:0> trap(:TERM) { Logger.new("stdout").info "EventMonitor#start: ignoring SIGTERM" }
=> "DEFAULT"
irb(main):003:0> `kill #{Process.pid}`
log writing failed. can't be called from trap context
=> ""
```

Move this logic up to the caller of start and rescue a SignalException
with "SIGTERM" message instead.

See also: https://github.com/ManageIQ/manageiq/pull/2386

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_catcher/runner.rb
@@ -26,6 +26,12 @@ class ManageIQ::Providers::Redhat::InfraManager::EventCatcher::Runner < ManageIQ
       @queue.enq events
       sleep_poll_normal
     end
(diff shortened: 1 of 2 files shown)


jrafanie:
Actually, on second thought, this ignoring of SIGTERM is not needed and appears to have been copy/pasted from old code from 2011 and 2009, I'll open a PR to remove it.

jrafanie:
Opened #80 instead

PR #81, opened 2017-08-21
Add method to check if vm exists on provider

borod108:
The way it was done before is by checking for an exception when refreshing the vm.

This is part of a fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1478108

This is required for: https://github.com/ManageIQ/manageiq-automation_engine/pull/66


borod108:
@miq-bot assign @masayag

borod108:
@oourfali can we merge this?

PR #87, opened 2017-08-29
Highest allowed API version

jntullo:
the `raw_connect` method was using `highest_allowed_api_version`, an instance method. I attempted to make it into a class method and include it as an instance method, however, the `cache_key` method requires an id, which does not work for the instance method.

It appears that the highest allowed API version is 4, since there are only two `raw_connect` methods (v3 and v4). My thought was to store this in a constant for the use of `raw_connect` to simplify and not require the use of the `cache_key`, but I am open to other suggestions.

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -4,6 +4,8 @@ require 'resolv'
 module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
   extend ActiveSupport::Concern
 
+  HIGHEST_ALLOWED_API_VERSION = 4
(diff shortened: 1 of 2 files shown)


jntullo:
needed for https://github.com/manageiq/manageiq-ui-classic

jntullo:
needed for https://github.com/ManageIQ/manageiq-ui-classic/pull/1580

jntullo:
Required for https://github.com/ManageIQ/manageiq-ui-classic/pull/2243

oourfali:
@masayag can you please review?

masayag:
Based in merging https://github.com/ManageIQ/manageiq-providers-ovirt/pull/101 I'm closing the current PR.

PR #88, opened 2017-08-30
Implement 'update_cloud_init' for V4

jhernand:
Currently the provisioning of virtual machines from templates using
cloud-init fails when using version 4 of the API. The reason is that the
'update_cloud_init!' method used to set configure cloud-init isn't
implemented for that version of the API. This patch implements it,
preserving the logic used in the implementation done for V3, as we need
to make sure that existing customization scripts will continue working.

This patch is part of the fix for the following bug:

  Cloud-init not working when using api v4
  https://bugzilla.redhat.com/1486340

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
@@ -110,17 +110,17 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
     end
 
     def vm_boot_from_cdrom(operation, name)
-      operation.get_provider_destination.boot_from_cdrom(name)
+      operation.with_provider_destination { |d| d.boot_from_cdrom(name) }
     rescue Ovirt::VmNotReadyToBoot
       raise OvirtServices::VmNotReadyToBoot
     end
 
     def detach_floppy(operation)
-      operation.get_provider_destination.detach_floppy
+      operation.with_provider_destination(&:detach_floppy)
     end
 
     def vm_boot_from_network(operation)
-      operation.get_provider_destination.boot_from_network
+      operation.with_provider_destination(&:boot_from_network)
     rescue Ovirt::VmNotReadyToBoot
       raise OvirtServices::VmNotReadyToBoot
     end
--- a/app/models/manageiq/providers/redhat/infra_manager/provision.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision.rb
@@ -9,8 +9,8 @@ class ManageIQ::Providers::Redhat::InfraManager::Provision < MiqProvision
     "Vm"
   end
 
-  def get_provider_destination
-    return nil if destination.nil?
-    destination.with_provider_object { |rhevm_vm| return rhevm_vm }
+  def with_provider_destination
+    return if destination.nil?
+    destination.with_provider_object { |obj| yield obj }
   end
 end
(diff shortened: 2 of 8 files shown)


jhernand:
@borod108 @pkliczewski @masayag @oourfali Please review.

Note that I am marking this as `WIP` because the change is needed, but not enough to fix the bug. Once this method is implemented I am seeing the problem with closed connections, so the complete provisioning doesn't work yet.

jhernand:
[Bug 1486340 - Cloud-init not working when using api v4](https://trello.com/c/tEGSyZ6J/196-bug-1486340-cloud-init-not-working-when-using-api-v4)

jhernand:
I have added to this pull request an additional commit that replaces the `get_provider_destination` method with a new `with_provider_destination` method that does not _leak_ the provider object, and thus the connection. With that I was able to complete provisioning using V4. But when I try to do it with V3 it fails. Not sure if it is related. The problem that I see is that the `source.ext_management_system` attribute is `nil`. During the first refresh that attribute is populated correctly, but later, when doing incremental refreshes, it is lost.

pkliczewski:
This [1] PR should fix the `ems` being `nil` issue.

[1] https://github.com/ManageIQ/manageiq/pull/15924

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -110,17 +110,17 @@ def powered_on_in_provider?(vm)
...
     def vm_boot_from_cdrom(operation, name)
-      operation.get_provider_destination.boot_from_cdrom(name)
+      operation.with_provider_destination { |d| d.boot_from_cdrom(name) }
     rescue Ovirt::VmNotReadyToBoot
       raise OvirtServices::VmNotReadyToBoot
     end
 
     def detach_floppy(operation)
-      operation.get_provider_destination.detach_floppy
+      operation.with_provider_destination { |d| d.detach_floppy }
```
borod108: I think rubocop suggests you use:  operation.with_provider_destination(&:detach_floppy)

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -110,17 +110,17 @@ def powered_on_in_provider?(vm)
...
     end
 
     def vm_boot_from_network(operation)
-      operation.get_provider_destination.boot_from_network
+      operation.with_provider_destination { |d| d.boot_from_network }
```
borod108: I think rubocop suggests you use:  operation.with_provider_destination(&:boot_from_network)

app/models/manageiq/providers/redhat/infra_manager/provision.rb
```diff
@@ -9,8 +9,8 @@ def destination_type
     "Vm"
   end
 
-  def get_provider_destination
-    return nil if destination.nil?
-    destination.with_provider_object { |rhevm_vm| return rhevm_vm }
+  def with_provider_destination(&block)
+    return if destination.nil?
+    destination.with_provider_object { |obj| block.call(obj) }
```
borod108: I think codeclimet wants you to do:
destination.with_provider_object { |obj| yield obj }
and remove the &block argument.

http://www.rubydoc.info/gems/rubocop/0.47.1/RuboCop/Cop/Performance/RedundantBlockCall

jhernand:
Thanks @borod108, I updated as you suggested.

I also verified that with this pull request applied to the provider, and with the fix that @pkliczewski prepared for the core, I can successfully provision a VM using cloud-init, both with V3 and V4.

PR #91, opened 2017-09-06
Support sysprep for windows templates

pkliczewski:
This PR provide sysprep support by sending complete xml file so there is no need to further customize it on ovirt side.

Bug-Url:
https://bugzilla.redhat.com/1323289

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
@@ -171,7 +171,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
 
     def vm_start(vm, cloud_init)
       vm.with_provider_object do |rhevm_vm|
-        rhevm_vm.start { |action| action.use_cloud_init(true) if cloud_init }
+        rhevm_vm.start { |action| action.use_cloud_init(true) if cloud_init[:use_cloud_init] }
       end
     rescue Ovirt::VmAlreadyRunning
     end
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -229,10 +229,8 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       end
     end
 
-    def vm_start(vm, cloud_init)
-      opts = {}
+    def vm_start(vm, opts = {})
       vm.with_provider_object(VERSION_HASH) do |rhevm_vm|
-        opts = {:use_cloud_init => cloud_init} if cloud_init
         rhevm_vm.start(opts)
       end
     rescue OvirtSDK4::Error
@@ -457,6 +455,14 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
           :storage_domain => args[:storage] && {:id => args[:storage].id}
         }.compact
         OvirtSDK4::Template.new(options)
+      end
+
+      def update_sysprep!(content)
+        update(OvirtSDK4::Vm.new(
+                 :initialization => {
+                   :custom_script => content
+                 }
+        ))
       end
     end
 
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
@@ -19,6 +19,15 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration
     end
   end
 
+  def configure_sysprep
+    content = get_option(:sysprep_upload_text)
+
+    return unless content
+    with_provider_destination { |d| d.update_sysprep!(content) }
+
+    phase_context[:boot_with_sysprep] = true
+  end
+
   def configure_container
     vm.with_provider_object(:version => vm.ext_management_system.highest_allowed_api_version) do |rhevm_vm|
       configure_container_description(rhevm_vm)
@@ -27,7 +36,13 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration
       configure_cpu(rhevm_vm)
       configure_host_affinity(rhevm_vm)
       configure_network_adapters
-      configure_cloud_init
+
+      sysprep_option = get_option(:sysprep_enabled)
+      if sysprep_option == 'file'
+        configure_sysprep
+      elsif sysprep_option == 'fields'
+        configure_cloud_init
+      end
     end
   end
 
--- a/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
@@ -55,14 +55,29 @@ class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisio
 
   def allowed_customization_templates(options = {})
     if supports_native_clone?
-      return allowed_cloud_init_customization_templates(options)
+      allowed_cloud_init_customization_templates(options)
     else
-      return super(options)
+      super(options)
     end
   end
 
   def allowed_datacenters(_options = {})
     super.slice(datacenter_by_vm.try(:id))
+  end
+
+  def allowed_customization(_options = {})
+    src = get_source_and_targets
+    return {} if src.blank?
+    return {"fields" => "Specification"} if @values[:forced_sysprep_enabled] == 'fields'
+
+    result = {"disabled" => "<None>"}
+
+    case src[:vm].platform
+    when 'windows' then result["file"] = "Sysprep Answer File"
+    when 'linux'   then result["fields"] = "Specification"
+    end
+
+    result
   end
 
   def datacenter_by_vm
--- a/spec/models/manageiq/providers/redhat/infra_manager/provision_workflow_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/provision_workflow_spec.rb
@@ -107,6 +107,7 @@ describe ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow do
 
   context "#allowed_customization_templates" do
     let(:workflow) { described_class.new({:src_vm_id => template.id}, admin) }
+    let(:source_vm) { double("OvirtSDK4::Vm") }
 
     it "should retrieve cloud-init templates when cloning" do
       options = {'key' => 'value'}
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/state_machine.rb
@@ -76,6 +76,7 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::StateMachine
 
   def autostart_destination
     destination.custom_attributes.create!(:name => "miq_provision_boot_with_cloud_init") if phase_context[:boot_with_cloud_init]
+    destination.custom_attributes.create!(:name => "miq_provision_boot_with_sysyprep") if phase_context[:boot_with_sysprep]
 
     super
   end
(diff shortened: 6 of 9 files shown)


pkliczewski:
@borod108 @jhernand Please take a look

pkliczewski:
[Bug 1323289 - [RFE] Support sysprep for windows templates for rhev](https://trello.com/c/yJqC2Bd6/191-bug-1323289-rfe-support-sysprep-for-windows-templates-for-rhev)

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -405,6 +405,19 @@ def update_cloud_init!(content)
           )
         )
       end
+
+      def platform
+        vm = get
+        vm.os.type
+      end
```
jhernand: I wonder if we have a better way to determine if a virtual machine has Windows as the operating system. ManageIQ does extract that information from the virtual machine, and also performs smart analysis. Isn't there a way to use that information in order to determine if the OS is Windows in a more reliable way? I mean, the `os.type` may change, it isn't part of the contract of the oVirt API, it comes from configuration files in the engine. Tomorrow it could be `win` instead of `windows` (very unlikely) and then this would not work.

jhernand: Also I'd suggest to repeat the `windows` string, make it a constant, or a configuration setting.

pkliczewski: Good point, I will explore it and make `windows` a constant.

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -27,7 +35,12 @@ def configure_container
       configure_cpu(rhevm_vm)
       configure_host_affinity(rhevm_vm)
       configure_network_adapters
-      configure_cloud_init
+
+      if rhevm_vm.platform.downcase.include?("windows")
```
michalskrivanek: could it be selected on the user side? there is a chance that eventually cloud-init will work on windows too (it actually may already, if you provide the right file for it - we just don't do that in RHV)

pkliczewski: I can fetch both sysprep and cloud-init templates. User could decide what to choose but if we do not support it on the engine side I think it could be confusing for the user.

michalskrivanek: we do not have a template, but we do support the same generic payload mechanism to supply files in a cdrom image. So if miq provide the right template it may work. Not sure if it is supported for other providers yet, but in general there is a desire to support cloud-init on windows too

pkliczewski: I understand but to add cloud-init for windows is quite easy from miq perspective. I am not sure whether it is correct order of implementing it. We should first have it in the engine and later expose it in miq.

Does it mean that a customization template available in miq is good enough for the engine to consume?

michalskrivanek: not sure we'll be investing into it in rhv any time soon(there is an rfe, but it's not targeted). 

i have no idea if that customization template works, it's simply different than the one we have in rhv. the mechanism is mostly passthrough, so it's not like it matters to rhv what you pass in. But whether it works in all windows versions it's hard to tell. We did have(and still do) quite a lot of issues across various versions
The only way how you can "delegate" this responsibility to rhv would be not to use custom-supplied templates but rely only on parameters exposed by rhv - downside is that it limits the features

jhernand: I think that we should avoid delegating this to RHV is something we should avoid, if possible. Instead we should pass directly to RHV whatever the ManageIQ, so it will be in the hands of the ManageIQ user to assemble the right Unatend.xml (or whatever) file for his particular version of Windows. Otherwise whenever we have problems in RHV they will be automatically be problems in ManageIQ as well.

pkliczewski: I would not like to limit it but let the user to pass anything it is needed. Maybe we should provide rhv specific template for the user which could be customized and passed to rhv.

michalskrivanek: that is fine with me, it's just important to acknowledge the difference between what RHV supports and what this PR adds to MIQ - that it doesn't cover the exact same feature set.

pkliczewski: Do you mean to update PR description to state it clearly what are the differences?

pkliczewski:
This PR requires https://github.com/ManageIQ/manageiq/pull/16071

pkliczewski:
@masayag @jhernand please review

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -171,7 +171,7 @@ def start_clone(source, clone_options, phase_context)
 
     def vm_start(vm, cloud_init)
       vm.with_provider_object do |rhevm_vm|
-        rhevm_vm.start { |action| action.use_cloud_init(true) if cloud_init }
+        rhevm_vm.start { |action| action.use_cloud_init(true) unless cloud_init.empty? }
```
bdunne: You can probably use `if cloud_init.present?`

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -458,6 +456,19 @@ def build_template_from_hash(args)
         }.compact
         OvirtSDK4::Template.new(options)
       end
+
+      def platform
+        vm = get
+        vm.os.type
+      end
+
+      def update_sysprep!(content)
```
bdunne: Will this `raise` if it fails?

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
bdunne: I think could-init or sysprep is fine for this PR.  In theory, I could see people wanting to do kickstart and cloud-init as well, but I don't think our dialogs are set up to allow selection of multiple templates yet.  This could be a follow-up set of PR's cc @gmcculloug

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -54,13 +54,33 @@ def update_field_visibility_linked_clone(_options = {}, f)
   end
 
   def allowed_customization_templates(options = {})
-    if supports_native_clone?
+    source_vm = get_source_vm
+    source = load_ar_obj(source_vm)
+
+    return [] if source.blank?
+    if source_vm.platform.downcase.include?("windows")
+      return allowed_sysprep_customization_templates(source)
```
bdunne: The `return` is unnecessary since there is nothing after the conditional.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
pkliczewski: OK

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
pkliczewski: It could raise sdk communication error. Will fix.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
pkliczewski: Done

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
pkliczewski: The error will be propagated to the user so it is correct way of "not" handling it here.

spec/models/manageiq/providers/redhat/infra_manager/provision_workflow_spec.rb
```diff
@@ -115,6 +116,15 @@
       workflow.allowed_customization_templates(options)
     end
 
+    it "should retrieve sysprep templates when cloning" do
+      options = {'key' => 'value'}
+      allow(source_vm).to receive(:platform).and_return('windows')
+      allow(workflow).to receive(:get_source_vm).and_return(source_vm)
+
+      expect(workflow).to receive(:allowed_sysprep_customization_templates).with(source_vm)
+      workflow.allowed_customization_templates(options)
```
bdunne: This isn't testing the logic in `allowed_customization_templates`

pkliczewski:
@jhernand @bdunne @michalskrivanek please review one more time

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
```diff
@@ -171,7 +171,7 @@ def start_clone(source, clone_options, phase_context)
 
     def vm_start(vm, cloud_init)
       vm.with_provider_object do |rhevm_vm|
-        rhevm_vm.start { |action| action.use_cloud_init(true) if cloud_init }
+        rhevm_vm.start { |action| action.use_cloud_init(true) if cloud_init.present? }
```
bdunne: Does this have the same caller as the method in the v4 strategy?  If so, the parameters need updating

pkliczewski: V3 do not support sysprep. The code is different.

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -27,7 +36,13 @@ def configure_container
       configure_cpu(rhevm_vm)
       configure_host_affinity(rhevm_vm)
       configure_network_adapters
-      configure_cloud_init
+
+      sysprep_option = get_option(:sysprep_enabled)
+      if sysprep_option == 'file'
+        configure_sysprep
+      elsif sysprep_option == 'fields'
```
jhernand: So we will only configure cloud-init if sysprep_enabled == fields? Sounds strange to me. What would happen if the user tries to use cloud-init without setting any sysprep-related option?

pkliczewski: Now we have a choice in the UI. We can go without a configuration or we need to use choose from the UI `specification` option and we can see other fields on the UI. It is not possible to use it without it.
I did not choose a name and I agree it is not that meaningful but this is how it works for vmware.

jhernand: So, if I understand correctly, the user will have an additional step in the UI in order to use cloud-init: to choose the "fields" option. After that the UI for cloud-init will be like it used to be? Or will it also require the user to upload a cloud-init configuration file? Can you maybe add some screenshots showing how it looks like now? Both for sysprep and cloud-init?

jhernand: Note that I don't object to that additional step, just want to make sure that the rest of the UI is like it used to be for cloud-init.

pkliczewski: UI structure and naming is exactly the same.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v3.rb
bdunne: I guess my concern is that the v3 call is `vm_start(vm, Boolean)` and the v4 call is `vm_start(vm, Hash)`, so the caller needs to be aware of this if it's shared code.

pkliczewski: Please check here https://github.com/ManageIQ/manageiq-providers-ovirt/pull/91/files#diff-7f31c34b148696c2c255077a322e8e66R10. I changed it to have vm_start(vm, Hash) for both.

pkliczewski:
@oourfali @masayag Are we OK to take it?

PR #92, opened 2017-09-07
Add connection manager

jhernand:
This patch adds a new `ConnectionManager` singleton intended to simplify
re-use of connections in single threaded processes. This class will
manage a set of connections, indexed by the identifier of the EMS. The 
connections will be created when needed, re-created when options change,
and closed when the worker finishes or when the EMS is removed from the 
database.
    
Checking for removed EMSs is performed periodically, with an interval
controlled by the new `connection_manager.purge_interval` parameter. The 
default value is 1 hour:

```yaml
:ems:
  :ems_ovirt:
    :connection_manager:
      :purge_interval: 1.hour
```
    
The existing code is also changed so that it uses the connection manager
instead of creating the connections directly, and so that it does not 
close the connections explicitly.
    
This only applies to V4, the V3 code does not change.

### DIFF
--- a/app/models/manageiq/providers/redhat/connection_manager.rb
+++ b/app/models/manageiq/providers/redhat/connection_manager.rb
@@ -1,0 +1,226 @@
+#
+# Copyright (c) 2017 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'ovirtsdk4'
+
+#
+# This class is a responsible for managing a set of `OvirtSDK4::Connection` objects. Each connection is
+# identified by the `id` attribute of the corresponding EMS, or by `nil` if there is no such EMS created yet.
+#
+# The connections will be created the first time that they are needed.
+#
+# Connections will be closed and created again if any of the options used to create them change.
+#
+# All the connections will be closed when the process finishes.
+#
+# Periodically the manager will obtain the identifiers of the EMSs from the database, and will automatically
+# close the connections that correspond to identifiers that no longer exist.
+#
+class ManageIQ::Providers::Redhat::ConnectionManager
+  #
+  # Returns the singleton instance.
+  #
+  def self.instance
+    @instance ||= new
+  end
+
+  #
+  # Creates a connection manager with an empty set of connections. This class is intended to be used as a
+  # singleton, so use the `instance` class method instead.
+  #
+  # @api private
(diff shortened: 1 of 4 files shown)


jhernand:
@pkliczewski @borod108 @masayag please review.

jhernand:
[Use a connection pool to manage the connections from the provider to the engine. We need to make sure that connections are re-used as much as possible, and they are closed when no longer needed.](https://trello.com/c/CmcKCbgC/195-use-a-connection-pool-to-manage-the-connections-from-the-provider-to-the-engine-we-need-to-make-sure-that-connections-are-re-use)

jhernand:
@borod108 @pkliczewski @masayag I have changed the implementation significantly in the new version of the pull request. Basically it is now assuming that the connection key is the identifier of the EMS, and taking advantage of that to be able to periodically query for EMS ids and close the corresponding connections.

So connections are now closed when the parameters change, when the process finishes, and when the EMS is removed (after a certain interval, by default one hour).

Please review again.

app/models/manageiq/providers/redhat/connection_manager.rb
```diff
@@ -0,0 +1,227 @@
...
+  # credentials, for example changes of user names or passwords.
+  #
+  # @param id [Object] The id of the EMS.
+  #
+  # @param opts [Hash] The options that will be used to create the connection if it doesn't exist yet. The possible
+  #   values are the same used in the constructor of the `OvirtSDK4::Connection` object.
+  #
+  # @return [OvirtSDK4::Connection] The connection that matches the given EMS id and options.
+  #
+  def get(id, opts)
```
pkliczewski: Shouldn't we validate whether the id represents existing ems?

app/models/manageiq/providers/redhat/connection_manager.rb
```diff
@@ -0,0 +1,227 @@
...
+    entry = @registry[id]
+    return entry.connection if entry && entry.compatible?(opts)
+
+    # If there is an entry but it isn't compatible, then close and remove it:
+    if entry
+      $rhevm_log.info(
+        "Existing connection for EMS with identifier '#{id}' and URL '#{entry.options[:url]}' isn't compatible " \
+        "with the requested options, will close it and create a new one."
+      )
+      entry.connection.close
```
pkliczewski: why not to use close(id) method here?

jhernand: Yes, indeed, I will change.

app/models/manageiq/providers/redhat/connection_manager.rb
jhernand: I think that isn't necessary. In addition the id may be `nil` if the EMS hasn't been created yet (when testing connection details).

jhernand:
Addressed the Rubocop complaints and changed to use the `close` method as suggested by @pkliczewski.

pkliczewski:
Looks good to me

oourfali:
@jhernand do we want this in, or should we do further testing?

jhernand:
@oourfali I would like to run automation tests on this before merging it. I already applied the changes to one of the QE environments and asked the QE team to run those automation tests. I will confirm here when those tests are completed, or when the pull request should otherwise be merged.

jhernand:
The automation tests passed, please merge.

PR #95, opened 2017-09-14
Support publish VM

masayag:
Creating a template from a VM is supported by RHV since forever.
The PR introduces a new dialog to allow the creation of a template from
a VM, including the ability to seal a template (for non-windows vms).

https://bugzilla.redhat.com/show_bug.cgi?id=1373076

**Screenshots of the new dialog:**
Request tab:
publish_vm_request_tab

Catalog tab:
publish_vm_catalog_tab

Environment tab:
publish_vm_environment_tab

Schedule tab:
publish_vm_schedule_tab

Due to a limitation of RHV, when specifying a storage domain, all of the template's disks will be created on the specified storage domain (while on RHV the user can specify storage domain per disk)

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -36,11 +36,28 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       phase_context = args[:phase_context]
       logger = args[:logger]
 
+      if source.template?
+        vm_clone_completed?(logger, phase_context, source)
+      else
+        template_clone_completed?(logger, phase_context, source)
+      end
+    end
+
+    def vm_clone_completed?(logger, phase_context, source)
       source.with_provider_connection(VERSION_HASH) do |connection|
         vm = vm_service_by_href(phase_context[:new_vm_ems_ref], connection).get
         status = vm.status
         logger.info("The Vm being cloned is #{status}")
         status == OvirtSDK4::VmStatus::DOWN
+      end
+    end
+
+    def template_clone_completed?(logger, phase_context, source)
+      source.with_provider_connection(VERSION_HASH) do |connection|
+        template = template_service_by_href(phase_context[:new_vm_ems_ref], connection).get
+        status = template.status
+        logger.info("The status of the template being cloned is #{status}")
+        status == OvirtSDK4::TemplateStatus::OK
       end
     end
 
@@ -170,7 +187,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
     end
 
     def get_vm_proxy(vm, connection)
-      VmProxyDecorator.new(connection.system_service.vms_service.vm_service(vm.uid_ems), self)
+      VmProxyDecorator.new(connection.system_service.vms_service.vm_service(vm.uid_ems), connection, self)
     end
 
     def collect_disks_by_hrefs(disks)
@@ -202,6 +219,13 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       source.with_provider_object(VERSION_HASH) do |rhevm_template|
         vm = rhevm_template.create_vm(clone_options)
         populate_phase_context(phase_context, vm)
+      end
+    end
+
+    def make_template(source, clone_options, phase_context)
+      source.with_provider_object(VERSION_HASH) do |rhevm_vm|
+        template = rhevm_vm.make_template(clone_options)
+        populate_phase_context(phase_context, template)
       end
     end
 
@@ -298,11 +322,13 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
     end
 
     class VmProxyDecorator < SimpleDelegator
-      attr_reader :service
-      def initialize(vm, service)
-        @obj = vm
-        @service = service
-        super(vm)
+      attr_reader :ovirt_services, :connection
+
+      def initialize(vm_service, connection, ovirt_services)
+        @obj = vm_service
+        @connection = connection
+        @ovirt_services = ovirt_services
+        super(vm_service)
       end
 
       def update_memory_reserve!(memory_reserve_size)
@@ -326,7 +352,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
 
       def update_host_affinity!(dest_host_ems_ref)
         vm = get
-        vm.placement_policy.hosts = [OvirtSDK4::Host.new(:id => service.uuid_from_href(dest_host_ems_ref))]
+        vm.placement_policy.hosts = [OvirtSDK4::Host.new(:id => ovirt_services.uuid_from_href(dest_host_ems_ref))]
         update(vm)
       end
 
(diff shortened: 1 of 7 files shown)


masayag:
[Bug 1373076 - [RFE] Publish VM to a template](https://trello.com/c/4Xi2sxTL/185-bug-1373076-rfe-publish-vm-to-a-template)

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -44,6 +52,15 @@ def clone_completed?(args)
       end
     end
 
+    def template_clone_completed?(logger, phase_context, source)
+      source.with_provider_connection(VERSION_HASH) do |connection|
+        template = template_service_by_href(phase_context[:new_vm_ems_ref], connection).get
+        status = template.status
+        logger.info("The Template being cloned is #{status}")
```
pkliczewski: Shouldn't it be "Cloned template status is #{status}"?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -437,6 +490,10 @@ def cluster_from_href(href, connection)
       connection.system_service.clusters_service.cluster_service(uuid_from_href(href)).get
     end
 
+    def storage_from_href(href, connection)
```
pkliczewski: We already fetch storage domain in different [place](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb#L72). I think that we start to have similar responsibilities in `inventory` and `ovirt_services`. In my opinion there is no need to have two distinct object having the same responsibility.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -476,6 +533,11 @@ def vm_service_by_href(href, connection)
       connection.system_service.vms_service.vm_service(vm_uuid)
     end
 
+    def template_service_by_href(href, connection)
```
pkliczewski: Same [here](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb#L110)

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
```diff
@@ -10,13 +10,29 @@ def destination_image_locked?
...
   def prepare_for_clone_task
     raise MiqException::MiqProvisionError, "Provision Request's Destination VM Name=[#{dest_name}] cannot be blank" if dest_name.blank?
     raise MiqException::MiqProvisionError, "A VM with name: [#{dest_name}] already exists" if source.ext_management_system.vms.where(:name => dest_name).any?
 
+    prepare_clone_options
+  end
+
+  def prepare_clone_options
+    if source.template?
+      vm_clone_options
```
pkliczewski: Please add some comment here. I was puzzled by this code because I see `is_templare` => `true` then do vm. I noticed similar code in line 80 where comment helped.

app/models/manageiq/providers/redhat/infra_manager/provision/state_machine.rb
```diff
@@ -33,15 +33,19 @@ def poll_clone_complete
...
   def customize_destination
-    if destination_image_locked?
-      _log.info("Destination image locked; re-queuing")
-      requeue_phase
+    if source.template?
+      if destination_image_locked?
+        _log.info("Destination image locked; re-queuing")
+        requeue_phase
+      else
+        message = "Starting New #{destination_type} Customization"
```
pkliczewski: Why we use capital letter for "new" and "customization"?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -135,4 +135,13 @@ def validate_memory_limit(_field, values, dlg, fld, _value)
       _("%{description} VM Memory is larger than Memory Limit") % {:description => required_description(dlg, fld)}
     end
   end
+
+  def validate_seal_template(_field, values, _dlg, _fld, _value)
+    seal = get_value(values[:seal])
+    return nil unless seal
+
+    if get_source_vm.platform == 'windows'
+      _("Template Sealing is supported only for non-Windows OS.")
```
pkliczewski: Is there a reason to have "sealing" starting with capital letter?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
masayag: we'd might access RHV from different places of the provider. We can think of introducing a class that will be responsible for interacting between the provider to RHV via v4, I'm not sure this is the PR to introduce such.

We should first get into agreement if we'd prefer that method over letting each workflow/scenario interact independently using the engine's connection.

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
masayag: agreed. I've start looking into the 'clone vm' workflow and encountered that using the `source.template?` will have to be replaced by a different method to identify the existing flow.

With the current PR, we're reusing the same workflow entity for add vm and add template. However, in 'clone vm' both source and destination are vms. I'll handle it as part of 'clone vm' PR.

I'll add comment here as well.

app/models/manageiq/providers/redhat/infra_manager/provision/state_machine.rb
masayag: developer excuses #211: "it was like that before"

seems like other providers (vmware and microsoft) have the exact message (with the capitals)

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
masayag: fixed.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
pkliczewski: I am not saying that we should fix it in this PR. I am saying that we already have similar code in different place. I think it would be better to define responsibilities of the classes we use to talk to the engine.

pkliczewski:
@masayag code climate noticed duplication. It seems not that hard to have one method. What do you think?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
masayag: We'll track refactoring of ovirt_services and querying of engine's request by issue https://github.com/ManageIQ/manageiq-providers-ovirt/issues/100

pkliczewski: Thanks

pkliczewski:
Based on my conversation with @masayag we will ignore codeclimate for now.

PR #96, opened 2017-09-19
Target new template when using api v4

pkliczewski:
Original implementation only provided such functionality for api v3.


pkliczewski:
@masayag @borod108 Please review

masayag:
Verified it also as part of 'publish vm' flow. The template's event is being caught instantly and triggers the template refresh.

pkliczewski:
Are we OK to have it in?

masayag:
Yes, the refresh error seems not related to this fix, and this fix is required for the 'publish vm' to work properly.

PR #97, opened 2017-09-19
Parse the serial number during refresh

borod108:
Parse the serial number of host during refresh.
This is required for:
https://bugzilla.redhat.com/show_bug.cgi?id=1410183

Required for:
https://github.com/ManageIQ/manageiq/pull/15992

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory_collection_default/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory_collection_default/infra_manager.rb
@@ -479,7 +479,8 @@ class ManageIQ::Providers::Redhat::InventoryCollectionDefault::InfraManager < Ma
           :memory_mb,
           :model,
           :networks,
-          :number_of_nics
+          :number_of_nics,
+          :serial_number
         ]
       }
     end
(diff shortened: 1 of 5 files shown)


borod108:
@durandom who should I ask to review this?

pkliczewski:
Do we need to update any specs for this change? Where is corresponding change for graph refresh?

borod108:
@pkliczewski you are right need to change graph refresh as well, can you point me to the place exactly?
The specs will change once the async is in. I do not want to rewrite them twice.

pkliczewski:
@borod108 I think you need to update [here](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb#L205)

durandom:
@borod108 afaik the schema pr is already merged, so nobody from the core needs to review this. Any corresponding PR for the UI or API?

borod108:
@pkliczewski done the changes, can you review again please?

borod108:
@pkliczewski can you please review? I added the fixes you requested.

borod108:
@oourfali @masayag can you please merge?

app/models/manageiq/providers/redhat/inventory_collection_default/infra_manager.rb
```diff
@@ -478,6 +478,7 @@ def hardwares_attributes
           :manufacturer,
           :memory_mb,
           :model,
+          :serial_number,
```
bdunne: Can these be alphabetical?

borod108: done

PR #101, opened 2017-10-03
Require Ovirt in ApiIntegration

jntullo:
Resolves uninitialized constant error described in https://github.com/ManageIQ/manageiq-providers-ovirt/pull/67

@miq-bot assign @borod108 
@miq-bot add_label bug


jntullo:
@borod108 Thanks for the suggestion! That did the trick 👍

PR #106, opened 2017-10-13
Avoid NoMethod error in TemplatePreloadedAttributesDecorator.new

jhernand:
Currently the constructor of the `TemplatePreloadedAttributesDecorator`
class receives a parameter that defaults to 'nil', and that it is
actually never used. But the implementation of the constructor tries
call methods on it, `NoMethodError`. The net result is that refresh
fails. To avoid that this patch removes that unused parameter.

This patch addresses the following issue:

  RHV Refresh worker throwing error in evm.log
  https://bugzilla.redhat.com/1501162


jhernand:
@borod108 please review.

borod108:
@miq-bot assign @oourfali

PR #107, opened 2017-10-16
Require oVirt Ruby SDK 4.1.9 or newer

jhernand:
This patch changes the provider so that it directly requires version
4.1.9 or newer of the oVirt Ruby SDK.

The new version of the SDK addresses a problem when using the same
connection from multiple threads, described in the following bug:

  Access to RHV using the oVirt SDK may crash the events worker
  https://bugzilla.redhat.com/1496848


jhernand:
@pkliczewski @borod108 please review.

PR #108, opened 2017-10-18
Assume 3 when can't determine API version

jhernand:
Currently when using version 4 of the SDK is enabled the provider raises
an exception when it can't determine what versions of the API are
supported by the oVirt server. But this is a normal situation. It
happens in general when the credentials of the server are not known yet,
for example when the oVirt server has been discovered scanning a range
of IP addresses. If an exception is reaised then the discovery process
will fail. To avoid this issue this patch changes the provider so that
it tolerates this situation, and assumes that the server supports at
least version 3 of the API.

This patch addresses the following bug:

  RHV41 Provider Discovery failure
  https://bugzilla.redhat.com/1502963

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -85,8 +85,12 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
     probe_results = OvirtSDK4::Probe.probe(probe_args)
     probe_results.map(&:version) if probe_results
   rescue => error
-    _log.error("Error while probing supported api versions #{error}")
-    raise
+    # Note that errors when trying to find the supported API versions are perfectly normal, in particular authorization
+    # errors are expected, as in many situations, for example during discovery, the user name and the password aren't
+    # yet known. In these situations we *must* return an empty array, to indicate to the caller that it wasn't possible
+    # to determine the supported API versions.
+    _log.info("Can't determine the API versions supported by the server: #{error}")
+    []
   end
 
   def supports_the_api_version?(version)
@@ -226,7 +230,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
   end
 
   def highest_supported_api_version
-    supported_api_versions.sort.last
+    supported_api_versions.sort.last || '3'
   end
 
   def highest_allowed_api_version


jhernand:
@pkliczewski @borod108 please review.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -85,8 +85,12 @@ def supported_api_versions_from_sdk
...
     probe_results.map(&:version) if probe_results
   rescue => error
-    _log.error("Error while probing supported api versions #{error}")
-    raise
+    # Note that errors when trying to find the supported API versions are perfectly normal, in particular authorization
+    # errors are expected, as in many situations, for example during discovery, the user name and the password aren't
+    # yet known. In these situations we *must* return an empty array, to indicate to the caller that it wasn't possible
+    # to determine the supported API versions.
+    _log.info("Can't determine the API versions supported by the server: #{error}")
+    []
```
pkliczewski: we won't have information why the probing failed. In situations like wrong credentials we won't state what happened. Am I right?

jhernand: We won't have information on why determining the supported versions failed. But that is just one part of the process of checking the credentials. After determining the supported versions we send additional requests to check the credentials (if the credentials are available). Those requests will fail, and the error will be reported to the user. The net result is that the user will see a message in the UI saying that the user name or password isn't correct.

pkliczewski: OK, so if the user gets this information I am fine with the change.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -226,7 +230,9 @@ def with_version4_vm_service(vm)
   end
 
   def highest_supported_api_version
-    supported_api_versions.sort.last
+    versions = supported_api_versions
+    return '3' if versions.empty?
```
borod108: Maybe versions.sort.last || '3' would be shortwr

borod108: I mean supported_api_versions.sort.last || '3'

jhernand: Yes, that makes sense, thanks @borod108, I am changing it.

jhernand:
The new version of the pull request addresses the comments from @borod108.

PR #110, opened 2017-10-19
Running 'ensure_network provider' as a separate job

AlonaKaplan:
'ensure_network' has an api call to ovirt, therefore, it should be run
as a separate jobs to avoid calling it directly from the ui.

All the tests that had to be fixed because of this api call were reverted.

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -22,7 +22,16 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
   supports :provisioning
   supports :refresh_new_target
 
-  before_save :ensure_managers
+  def ensure_managers
+    return unless enabled
+    ensure_network_manager
+    if network_manager
+      network_manager.name = "#{name} Network Manager"
+      network_manager.zone_id = zone_id
+      network_manager.provider_region = provider_region
+      network_manager.save!
+    end
+  end
 
   def ensure_network_manager
     providers = ovirt_services.collect_external_network_providers
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_1_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_1_spec.rb
@@ -1,14 +1,13 @@
 describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
   before(:each) do
     guid, server, zone = EvmSpecHelper.create_guid_miq_server_zone
-    @ems = FactoryGirl.build(:ems_redhat_with_ensure_managers, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
+    @ems = FactoryGirl.create(:ems_redhat, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
                               :port => 8443)
     @ovirt_service = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies::V4
     allow_any_instance_of(@ovirt_service)
       .to receive(:collect_external_network_providers).and_return(load_response_mock_for('external_network_providers'))
     @ems.update_authentication(:default => {:userid => "admin@internal", :password => "123456"})
     allow(@ems).to receive(:supported_api_versions).and_return(%w(3 4))
-    @ems.save
     stub_settings_merge(:ems => { :ems_redhat => { :use_ovirt_engine_sdk => true } })
   end
 
@@ -159,7 +158,7 @@ describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
 
     expect(Relationship.count).to eq(32)
 
-    expect(MiqQueue.count).to eq(13)
+    expect(MiqQueue.count).to eq(14)
   end
 
   def assert_ems
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_spec.rb
@@ -1,7 +1,7 @@
 describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
   before(:each) do
     _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
-    @ems = FactoryGirl.build(:ems_redhat_with_ensure_managers, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
+    @ems = FactoryGirl.create(:ems_redhat, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
                               :port => 8443)
     @ovirt_service = ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies::V4
     allow_any_instance_of(@ovirt_service)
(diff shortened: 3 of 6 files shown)


app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -24,7 +24,21 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
 
   before_save :ensure_managers
 
+  def ensure_network_manager_queue()
```
agrare: Don't need the empty parens

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -24,7 +24,21 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
...
+      :class_name  => self.class.name,
+      :instance_id => id,
+      :method_name => 'create_or_update_network_manager',
+      :zone        => my_zone,
+      :args        => []
+    )
+  end
+
   def ensure_network_manager
+    ensure_network_manager_queue()
```
agrare: Same, no parens needed

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -24,7 +24,21 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
 
   before_save :ensure_managers
 
+  def ensure_network_manager_queue()
+    MiqQueue.put(
+      :class_name  => self.class.name,
+      :instance_id => id,
+      :method_name => 'create_or_update_network_manager',
+      :zone        => my_zone,
+      :args        => []
```
agrare: No need to pass empty args, it defaults to an empty array

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -48,6 +62,7 @@ def ensure_network_manager
 
       if network_manager
         populate_network_manager_connectivity(auth_url)
+        network_manager.save
```
agrare: Can you use `.save!` so this will raise an exception and be logged if saving the network manager fails?

pazinio:
Hi @agarare. This pull request doesn't work well in all the cases. It works well in case the provider already exists. But if I'm in the create flow it doesn't work.
I get - MIQ(MiqQueue#deliver) Message id: [8806], Error: [undefined method `create_or_update_network_manager' for #<Class:0x00000009bb18a0>]
It seems that the infra_manger object is not fully initialized yet. 
It ends up with the workers of the network manager not started until full refresh of the infra provider. 
Do you have an idea how it can be solved?

AlonaKaplan:
Hi @agarare. This pull request doesn't work well in all the cases. It works well in case the provider already exists. But if I'm in the create flow it doesn't work.
I get - MIQ(MiqQueue#deliver) Message id: [8806], Error: [undefined method `create_or_update_network_manager' for #Class:0x00000009bb18a0]
It seems that the infra_manger object is not fully initialized yet.
It ends up with the workers of the network manager not started until full refresh of the infra provider.
Do you have an idea how it can be solved

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_target_vm_4_spec.rb
```diff
@@ -3,16 +3,11 @@
 
   before(:each) do
     _, _, zone = EvmSpecHelper.create_guid_miq_server_zone
-    @ems = FactoryGirl.build(:ems_redhat_with_ensure_managers, :zone => zone, :hostname => "192.168.1.105", :ipaddress => "192.168.1.105",
+    @ems = FactoryGirl.create(:ems_redhat, :zone => zone, :hostname => "192.168.1.105", :ipaddress => "192.168.1.105",
```
borod108: Why is this change required?

agrare:
> I get - MIQ(MiqQueue#deliver) Message id: [8806], Error: [undefined method `create_or_update_network_manager' for #Class:0x00000009bb18a0]
It seems that the infra_manger object is not fully initialized yet.
It ends up with the workers of the network manager not started until full refresh of the infra provider.
Do you have an idea how it can be solved

I got the same thing and its pretty obvious why once you see it haha
`[----] I, [2017-10-23T09:06:47.519370 #16361:2b090b533efc]  INFO -- : MIQ(MiqQueue.put) Message id: [36],  id: [], Zone: [default], Role: [], Server: [], Ident: [generic], Target id: [], Instance id: [], Timeout: [600], Priority: [100], State: [ready], Deliver On: [], Data: [], Args: []`

It is called from before_save, there is no `#id` yet

agrare:
I was talking to @borod108 on gitter and what if we did this, instead of doing this in the before_save/after_save callback, what if we did the check during refresh?  That way if a customer added a network endpoint after they created their InfraManager they don't have to delete&&create the whole provider to pick it up?

AlonaKaplan:
Yes, that is the approach I'm trying now.

AlonaKaplan:
@agrare @Ladas @masayag @borod108 please review

app/models/manageiq/providers/redhat/infra_manager.rb
```diff
@@ -1,3 +1,5 @@
+require 'thread'
```
agrare: What did you need from `thread`?

AlonaKaplan:
No need, leftover. Will remove.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_1_spec.rb
```diff
@@ -1,14 +1,13 @@
 describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
   before(:each) do
     guid, server, zone = EvmSpecHelper.create_guid_miq_server_zone
-    @ems = FactoryGirl.build(:ems_redhat_with_ensure_managers, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
+    @ems = FactoryGirl.create(:ems_redhat, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
```
borod108: is this change still required?

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_spec.rb
```diff
@@ -1,15 +1,14 @@
 describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
   before(:each) do
     _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
-    @ems = FactoryGirl.build(:ems_redhat_with_ensure_managers, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
+    @ems = FactoryGirl.create(:ems_redhat, :zone => zone, :hostname => "localhost", :ipaddress => "localhost",
```
borod108: same question

AlonaKaplan: Previously I changed it to 'build' since I had to mock the 'ovirt_service' result before the 'save', this is not needed any more, so I reverted it back to create.

AlonaKaplan:
depends on https://github.com/ManageIQ/manageiq/pull/16239

PR #111, opened 2017-10-23
[WIP] Move detecting the api versions to database query or worker

borod108:
The probing for api versions was done by connecting to the provider directly
which causes problems when done from a UI applience which has no access to the
provider.

We now use the "api_version" feild (which actually contains the product version)
to infer the api version supported. If it is empty we use Probing but now
we do it through the MiqQueue so it will run from a machine connected to the provider.


borod108:
after some discussion apperantely the queue part is not needed

PR #112, opened 2017-10-23
[WIP] Add method to generate UI URL

jhernand:
This patch adds a `generate_ui_url` method to the virtual machine class
that generates the complete URL needed to access the oVirt UI using the
credentials of the provider.


jhernand:
Replaced by ManageIQ/manageiq-providers-ovirt#133.

PR #114, opened 2017-10-24
Move detecting the api versions to database query

borod108:
The probing for api versions was done by connecting to the provider directly
which causes problems when done from a UI applience which has no access to the
provider.

We now use the "api_version" (which actually contains the product version)
to infer the api version supported. When the feild is not available we try probing.

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -4,6 +4,8 @@ require 'resolv'
 module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
   extend ActiveSupport::Concern
 
+  include SupportedApisMixin
+
   require 'ovirtsdk4'
   require 'ovirt'
 
@@ -61,43 +63,6 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
 
   def supports_port?
     true
-  end
-
-  def supported_api_versions
-    supported_api_versions_from_cache
-  end
-
-  def supported_api_versions_from_cache
-    cacher = Cacher.new(cache_key)
-    current_cache_val = cacher.read
-    force = current_cache_val.blank?
-    cacher.fetch_fresh(last_refresh_date, :force => force) { supported_api_versions_from_sdk }
-  end
-
-  def cache_key
-    "REDHAT_EMS_CACHE_KEY_#{id}"
-  end
-
-  def supported_api_versions_from_sdk
-    username = authentication_userid(:basic)
-    password = authentication_password(:basic)
-    probe_args = { :host => hostname, :port => port, :username => username, :password => password, :insecure => true }
-    probe_results = OvirtSDK4::Probe.probe(probe_args)
-    probe_results.map(&:version) if probe_results
-  rescue => error
-    # Note that errors when trying to find the supported API versions are perfectly normal, in particular authorization
-    # errors are expected, as in many situations, for example during discovery, the user name and the password aren't
-    # yet known. In these situations we *must* return an empty array, to indicate to the caller that it wasn't possible
-    # to determine the supported API versions.
-    _log.info("Can't determine the API versions supported by the server: #{error}")
-    []
-  end
-
-  def supports_the_api_version?(version)
-    if supported_api_versions.empty?
-      raise MiqException::MiqUnreachableError, "Not able to connect to the server."
-    end
-    supported_api_versions.map(&:to_s).include?(version.to_s)
   end
 
   def supported_auth_types
@@ -227,15 +192,6 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
     yield service
   ensure
     connection.close
-  end
-
-  def highest_supported_api_version
-    supported_api_versions.sort.last || '3'
-  end
-
-  def highest_allowed_api_version
-    return '3' unless use_ovirt_sdk?
-    highest_supported_api_version
   end
 
   def use_ovirt_sdk?
@@ -380,36 +336,6 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
     end
   end
 
-  class Cacher
-    attr_reader :key
-
-    def initialize(key)
-      @key = key
-    end
-
-    def fetch_fresh(last_refresh_time, options)
-      force = options[:force] || stale_cache?(last_refresh_time)
-      res = Rails.cache.fetch(key, :force => force) { build_entry { yield } }
-      res[:value]
-    end
-
-    def read
-      res = Rails.cache.read(key)
-      res && res[:value]
-    end
-
-    private
-
-    def build_entry
-      {:created_at => Time.now.utc, :value => yield}
-    end
-
-    def stale_cache?(last_refresh_time)
-      current_val = Rails.cache.read(key)
-      return true unless current_val && current_val[:created_at] && last_refresh_time
-      last_refresh_time > current_val[:created_at]
-    end
-  end
 
   private
 
--- a/app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
@@ -1,0 +1,62 @@
+module ManageIQ::Providers::Redhat::InfraManager::SupportedApisMixin
+  def supported_api_versions
+    return supported_api_versions_from_sdk(probe_args) if api_version.blank?
+    supported_api_versions_from_db
+  end
+
+  def highest_supported_api_version
+    supported_api_versions.sort.last || '3'
+  end
+
+  def highest_allowed_api_version
+    return '3' unless use_ovirt_sdk?
+    highest_supported_api_version
+  end
+
+  # This method is a result of api_version in the db not actually being the api_version
+  # it is actually the product version based on which we infer the api version.
+  # In the future I hope there will be a separate field created.
+  def supported_api_versions_from_db
+    key = product_version_to_api_version_regexps.keys.detect { |k| api_version =~ Regexp.new(k.to_s, Regexp::EXTENDED) }
+    product_version_to_api_version_regexps[key] || []
+  end
+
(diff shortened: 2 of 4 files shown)


borod108:
@masayag  @jhernand @pkliczewski if anyone can review this.
@agrare can you please see if it indeed solves the UI accessing the provider issue?

borod108:
@miq-bot remove-label wip

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
```diff
@@ -0,0 +1,85 @@
+module ManageIQ::Providers::Redhat::InfraManager::SupportedApisMixin
+  def supported_api_versions
+    return supported_api_versions_from_cache if api_version.blank?
+    supported_api_versions_from_db
+  end
+
+  DEFAULT_PRODUCT_VERSION_TO_API_VERSION_REGEX_HASH = {
+    '3\.\d+\.?\d*' => %w(3),
```
jhernand: There is no need to assume that the product version is numeric, or that it has only three components. I'd suggest to use `^3\..*$`. Note the `^` and `$` anchors to make sure that you are looking for the begin and end of the string, and not for some random position inside it. Same for version 4.

Also, I don't see the modification to the `settings.yaml` file. I think we should modify it, even if we have the defaults in the code. That way users can find the setting when taking a look at the `settings.yaml` file.

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
```diff
@@ -0,0 +1,85 @@
+module ManageIQ::Providers::Redhat::InfraManager::SupportedApisMixin
+  def supported_api_versions
+    return supported_api_versions_from_cache if api_version.blank?
```
jhernand: The cache is probably no longer convenient. It was convenient when determining the supported versions was expensive because it required talking to the server. But now that it only needs access to the database I think we can just remove completely the caching code.

spec/models/manageiq/providers/redhat/infra_manager_spec.rb
```diff
@@ -191,86 +191,114 @@
...
-            :created_at => Time.now.utc,
-            :value      => [3]
-          }
+      context "version 4.2.0-0.0.master.20170917124606.gita804ef7.el7.centos" do
+        let(:api_version) { "4.2.0-0.0.master.20170917124606.gita804ef7.el7.centos" }
+        it 'returns versions 3 and 4' do
+          expect(supported_api_versions).to match_array(%w(3 4))
         end
+      end
+    end
```
jhernand: Consider adding some tests to check product versions like `4.2.1.3`, `4.3`, or `4.2.1a`. I wouldn't bee surprised to see that in the future.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -63,36 +65,6 @@ def supports_port?
...
-  rescue => error
-    # Note that errors when trying to find the supported API versions are perfectly normal, in particular authorization
-    # errors are expected, as in many situations, for example during discovery, the user name and the password aren't
-    # yet known. In these situations we *must* return an empty array, to indicate to the caller that it wasn't possible
-    # to determine the supported API versions.
-    _log.info("Can't determine the API versions supported by the server: #{error}")
-    []
-  end
-
   def supports_the_api_version?(version)
```
masayag: why not to move other methods which deal with the supported api versions  ?

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
borod108: done

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
borod108: Regexps done.
I did not want to change the yaml because it is advanced stuff which i do not think a user should touch without really understanding what is going on, and in that case they should know how to add them. What do you think?

spec/models/manageiq/providers/redhat/infra_manager_spec.rb
borod108: done

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
jhernand: I think it is important to have it in the `settings.yaml` file. I, for example, have no idea of how to do that. If it was there already, I could probably modify it without having to know how to add it. So my suggestion is to add it.

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
```diff
@@ -0,0 +1,59 @@
...
+    highest_supported_api_version
+  end
+
+  # This method is a result of api_version in the db not actually being the api_version
+  # it is actually the product version based on which we infer the api version.
+  # In the future I hope there will be a separate field created.
+  def supported_api_versions_from_db
+    product_v_to_api_v_hash = DEFAULT_PRODUCT_VERSION_TO_API_VERSION_REGEX_HASH ||
+                              Setting::ems::ems_redhat.default_product_version_to_api_version_regex_hash
+    key = product_v_to_api_v_hash.keys.detect { |k| api_version =~ Regexp.new(k, Regexp::EXTENDED) }
```
jhernand: Does this guarantee that we will always select the first regex that matches, in the same order that they appear in the `settings.yaml` file? It is important that we make that predictable.

borod108: It does, although I do not think its a good thing to relay on while writing the regexps, I think they should be accurate enough to avoid confusion.

jhernand: I agree, both things: they should be accurate and they should be processed in a predictable order, same than in the `settings.yaml` file.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
borod108: done

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
borod108: Ok makes sense - I did, and I removed the default since it has no use in that case.

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
borod108: done :)

config/settings.yml
```diff
@@ -1,6 +1,9 @@
 ---
 :ems:
   :ems_redhat:
```
jhernand: Can this be under `ems_ovirt` ?

config/settings.yml
```diff
@@ -1,6 +1,9 @@
...
+    :default_product_version_to_api_version_regex_hash:
```
jhernand: I think this isn't the "default", rather the only way to set the mapping. I'd also remove the "regex" and "hash" parts from the name, as those are implementation details. So I suggest to name this `product_version_to_api_versions`.

config/settings.yml
```diff
@@ -1,6 +1,9 @@
...
+      :'^3\..*$': ['3']
+      :'^4\..*$': ['3', '4']
```
jhernand: Would it be possible to avoid the Ruby syntax in the values? I understand that it would require additional parsing in the client, but I'd say that is acceptable. Ideally this should be just the version numbers, separated by spaces or commas.

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
```diff
@@ -0,0 +1,53 @@
...
+  def highest_allowed_api_version
+    return '3' unless use_ovirt_sdk?
+    highest_supported_api_version
+  end
+
+  # This method is a result of api_version in the db not actually being the api_version
+  # it is actually the product version based on which we infer the api version.
+  # In the future I hope there will be a separate field created.
+  def supported_api_versions_from_db
+    product_v_to_api_v_hash = Settings::ems::ems_redhat.default_product_version_to_api_version_regex_hash
```
jhernand: Instead of accessing directly the settings, I think we could have a method to read it, and maybe do some parsing of the values, to avoid the Ruby syntax.

app/models/manageiq/providers/redhat/infra_manager/supported_apis_mixin.rb
```diff
@@ -0,0 +1,62 @@
...
+  # In the future I hope there will be a separate field created.
+  def supported_api_versions_from_db
+    key = product_version_to_api_version_regexps.keys.detect { |k| api_version =~ Regexp.new(k.to_s, Regexp::EXTENDED) }
+    product_version_to_api_version_regexps[key] || []
+  end
+
+  def product_version_to_api_version_regexps
+    return @parsed_regex_settings if @parsed_regex_settings
+    @parsed_regex_settings = {}
+    version_regex_settings = Settings::ems::ems_redhat.product_version_to_api_version_regexps
```
agrare: Should be `Settings.ems....` not `Settings::ems`

borod108:
@agrare Putting one or other version as default will not work well for this scenario, so we have to run through queue if we are running from the UI. But how to know if we are running from UI in general? Can we check somehow if we are running from a worker?

agrare:
@borod108 what I'm wondering is why we need to know the api version at all for `filter_allowed_hosts`?
Lets discuss elsewhere I don't want to clutter this PR, this LGTM

borod108:
@oourfali @masayag  seems everyone approved, can you merge please?

oourfali:
Merged. But please answer Adam

agrare:
@oourfali he pinged me on gitter we're discussing there :+1:

PR #117, opened 2017-10-26
We need to use lazy_find_by for hash index

Ladas:
We need to use lazy_find_by for hash index

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1506463


Ladas:
@pkliczewski I think for reproducing the bug, it's enough to have some GuestDevice addresses (probably missing in refresh spec mocked data)

PR #118, opened 2017-10-27
Network discovery probe relocation

gildub:
https://github.com/ManageIQ/manageiq-network_discovery/issues/8

Depends upon https://github.com/ManageIQ/manageiq/pull/16916

### DIFF
--- a/app/models/manageiq/providers/redhat/discovery.rb
+++ b/app/models/manageiq/providers/redhat/discovery.rb
@@ -1,0 +1,19 @@
+require 'manageiq/network/port'
+require 'ovirt'
+
+module ManageIQ
+  module Providers
+    module Redhat
+      class Discovery
+        def self.probe(ost)
+          Ovirt.logger = $rhevm_log if $rhevm_log
+
+          if ManageIQ::Network::Port.open?(ost, Ovirt::Service::DEFAULT_PORT) &&
+             Ovirt::Service.ovirt?(:server => ost.ipaddr, :verify_ssl => false)
+            ost.hypervisor << :rhevm
+          end
+        end
+      end
+    end
+  end
+end


lib/manageiq/providers/ovirt/redhat_rhevm_probe.rb
```diff
@@ -0,0 +1,16 @@
+require 'manageiq/network_discovery/port_scanner'
+
+module ManageIQ::Providers::Ovirt
+  class RedHatRhevmProbe
```
bdunne: Do we need the `RedHatRhevm` part?  I think the provider namespace would be enough

lib/manageiq/providers/ovirt/redhat_rhevm_probe.rb
```diff
@@ -0,0 +1,16 @@
...
+    def self.probe(ost)
+      require 'ovirt'
+      Ovirt.logger = $rhevm_log if $rhevm_log
+
+      if PortScanner.portOpen(ost, Ovirt::Service::DEFAULT_PORT)
+        if Ovirt::Service.ovirt?(:server => ost.ipaddr, :verify_ssl => false)
+          ost.hypervisor << :rhevm
```
bdunne: We may want to update to `:ovirt` in another PR.

bdunne: Does the `ost.hypervisor << :rhevm` belong in here?

I would expect to call `ManageIQ::Providers::Ovirt::Probe.probe(hostname or ip_address)` and have it return `true` or `false`.  Unless it could return many different things?

gildub: Sorry I'm not following you here.

gildub: Although I tend to agree with you, unfortunately the way probes are working so far is by passing `ost` structure which is updated by the probe themselves. Changing that is going to require a refactor across all probes, which is not the intend of this PR.

lib/manageiq/providers/ovirt/redhat_rhevm_probe.rb
gildub: Well we need a class name, because it's autoloaded in [1]. We could use `RedHatRhevm` unless you have another idea. 


[1] https://github.com/ManageIQ/manageiq-network_discovery/blob/master/lib/manageiq-network_discovery.rb#L6

app/models/manageiq/providers/redhat/discovery.rb
```diff
@@ -0,0 +1,20 @@
+require 'manageiq/network/port'
```
bdunne: This file doesn't exist yet.  Can you make that PR first?

app/models/manageiq/providers/redhat/discovery.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'ovirt'
+
+module ManageIQ
+  module Providers
+    module Redhat
+      class Discovery
+        def self.probe(ost)
+          Ovirt.logger = $rhevm_log if $rhevm_log
+
+          if Port.open?(ost, Ovirt::Service::DEFAULT_PORT)
```
bdunne: I think this needs to be fully qualified as `ManageIQ::Network::Port`.

lib/manageiq-providers-ovirt.rb
```diff
@@ -1,2 +1,10 @@
 require "manageiq/providers/ovirt/engine"
 require "manageiq/providers/ovirt/version"
+
+module ManageIQ
+  module Providers
+    module Redhat
+      autoload :Discovery, 'manageiq/providers/redhat/discovery'
```
bdunne: Do we need this?  The `app/models` directory is already autoloaded and it seems to work without this.

app/models/manageiq/providers/redhat/discovery.rb
gildub: This part is in https://github.com/ManageIQ/manageiq/pull/16916, I added the dependency in description.

app/models/manageiq/providers/redhat/discovery.rb
gildub: Done.

gildub:
@bdunne, when you have a chance could you please update review, thanks.

gildub:
@bdunne, flagging `gaprindashvili/yes` because the RFE is also targeting 5.9.1. I'm flagging the entire set of PRs that have the probes moved to manageiq/manageiq. I don't see any blocker or impact (with proper name space now), as long as all files are backported. Please let @simaishi and me know if you see an issue with this. Thanks

gildub:
@miq-bot add_label  gaprindashvili/yes

simaishi:

```
$ git log -1
commit f805eab20d54dee0c2cf9c9a157e07e64c6925fe
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Thu Feb 8 19:04:37 2018 -0500

    Merge pull request #118 from gildub/rhevm_probe
    
    Network discovery probe relocation
    (cherry picked from commit bdc63c17d8d9a9b71d0c7cf966294b2aaca500d3)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1552665
```

PR #127, opened 2017-11-02
Refresh a host when removed

pkliczewski:
Graph and regular refreshes fails when triggered by host removed event.
This PR fixes the issue.

Bug-Url:
https://bugzilla.redhat.com/1508831

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
+++ b/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
@@ -2,6 +2,19 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
   def initialize_inventory_collections
     add_targeted_inventory_collections
     add_remaining_inventory_collections([infra], :strategy => :local_db_find_references)
+    add_vms_dependency_collections(references(:vms))
+  end
+
+  def references(collection)
+    target.manager_refs_by_association.try(:[], collection).try(:[], :ems_ref).try(:to_a) || []
+  end
+
+  def infra
+    ManageIQ::Providers::Redhat::InventoryCollectionDefault::InfraManager
+  end
+
+  def add_vms_dependency_collections(manager_refs)
+    return if manager_refs.blank?
 
     add_inventory_collection(
       infra.ems_folder_children(
@@ -13,7 +26,6 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
         }
       )
     )
-
     add_inventory_collection(
       infra.ems_clusters_children(
         :dependency_attributes => {
@@ -22,7 +34,6 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
         }
       )
     )
-
     add_inventory_collection(
       infra.snapshot_parent(
         :dependency_attributes => {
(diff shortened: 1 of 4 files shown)


pkliczewski:
@Ladas @borod108 @masayag please review

pkliczewski:
[Bug 1508831 - Target refresh for host fail](https://trello.com/c/KwdHxX02/227-bug-1508831-target-refresh-for-host-fail)

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
```diff
@@ -117,6 +89,31 @@ def add_vms_inventory_collections(manager_refs)
         :strategy => :local_db_find_missing_references
       )
     )
+    add_inventory_collection(
+      infra.ems_folder_children(
+        :dependency_attributes => {
+          :clusters    => [collections[:ems_clusters]],
+          :datacenters => [collections[:datacenters]],
```
Ladas: placed here, this will result to nil and to the error 'no method .saved? for nil Class' we saw

pkliczewski: The issue is different. The code that I moved is related to vm only and it doesn't make sense when we target a host. I think that condition is not enough.

Ladas: right, what I mean is that collections[:clusters] is being add in add_clusters_inventory_collections, so it will be nil here. And the nil will result in error.

So the manual dependencies must be set when all of the collections have been initialized. You should be able to replace ```collections[:ems_clusters]``` with ```self.ems_clusters``` to get early exception

pkliczewski: In our model it is not possible to have a cluster or datacenter to be `nil`. We will always have at least one.

Ladas: right, but this is about

````
add_inventory_collection(infra.ems_clusters( ...
````

is creating the ```collections[:ems_clusters]``` so trying to access it before will result in nil.

That is why I propose using the ```self.ems_clusters```, you will get no method defined exception, until the ems_clusters InventoryCollection is initialized

Comment (Ladas) [reply app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb]: btw. that is the reason why the moved code was placed after all InventoryCollections were initialized (with the bug that add_remaining_inventory_collections is not initializing all the remaining, so we still get nil)

Comment (pkliczewski) [reply app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb]: Right, please see the new version. It should be ok now.

**Comment by oourfali** [comment]
Are the failures related?

**Comment by pkliczewski** [comment]
@oourfali The failure is codeclimete which shows similarity between two methods. In my opinion we are ok with ignoring it.

**Comment by pkliczewski** [comment]

**Comment by simaishi** [comment]

```
$ git log -1
commit a56769819cb512beecbeef4696b9ccb7cfcd4978
Date:   Fri Nov 3 17:57:10 2017 +0200

    Merge pull request #127 from pkliczewski/master
    
    Refresh a host when removed
    (cherry picked from commit 5371a675e68858e52b1a294d8a01fa142c72ee5e)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1510081
```

PR #132, opened 2017-11-06
Try both API versions in `raw_connect`

jhernand:
Currently the `raw_connect` method needs to receive the version of the
API in order to work correctly. If it doesn't receive it it fails
because the method that determines it just doesn't exist (it is an
instance method, but `raw_connect` is a class method). To avoid that
this patch changes the `raw_connect` method so that it tries with both
versions of the API.

This patch fixes partially the following bug:

  Failed validation when adding RHV provider
  https://bugzilla.redhat.com/1509432

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -41,8 +41,8 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
     # Starting with version 4 of oVirt authentication doesn't work when using directly the IP address, it requires
     # the fully qualified host name, so if we received an IP address we try to convert it into the corresponding
     # host name:
-    if resolve_ip_addresses?
-      resolved = resolve_ip_address(connect_options[:server])
+    if self.class.resolve_ip_addresses?
+      resolved = self.class.resolve_ip_address(connect_options[:server])
       if resolved != connect_options[:server]
         _log.info("IP address '#{connect_options[:server]}' has been resolved to host name '#{resolved}'.")
         default_endpoint.hostname = resolved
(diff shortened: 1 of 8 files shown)


jhernand:
@borod108 @masayag @pkliczewski @jntullo please review.

If/when this is merged we will be able to remove the `scheme` and `version` options from the call that the UI does to `raw_connect`.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -234,17 +234,88 @@ def process_api_features_support
...
+      v4_connection = nil
+      v4_error = nil
       begin
-        connection = public_send(connect_method, options)
-        connection.test(true)
+        v4_connection = raw_connect_v4(opts)
+        v4_connection.test(raise_exception=true)
+      rescue Exception => exception
+        v4_error = exception
+      end
```
pkliczewski: Why not to have ensure here as for v3 below? Do we want to keep it open?

pkliczewski: I saw the comment. Please ignore

pkliczewski:
@jhernand Can we fix codeclimate comments?

borod108:
@jhernand why can't we use probing with queue? something like https://github.com/ManageIQ/manageiq-providers-ovirt/pull/111/files#diff-4d8ab13ae69c2eade9d82675975c0defR56

jhernand:
@pkliczewski in the new version of the pull request I tried to address the style issues.

@borod108 probing via a queued request would mean two queued request, as the original call to `raw_connect` goes already via a queue. The result would be painfully slow.

jhernand:
[Bug 1509432 - Failed validation when adding RHV provider](https://trello.com/c/KiNKHJbL/235-bug-1509432-failed-validation-when-adding-rhv-provider)

jhernand:
@miq-bot add_label wip

I didn't test this with oVirt 3 yet, so please don't merge.

pkliczewski:
@jhernand Looks good to me, is it still WIP?

jhernand:
@pkliczewski yes, it is work in progress, I need to test that it works correctly with oVirt 3.6.

jhernand:
@miq-bot remove_label wip

Did some important changes to make this work with both versions of the API. I verified that it works correctly now for oVirt 4 and oVirt 3.

However, I also detected that the regression also affects the validation of the details of the C&U metrics endpoint. I am working on that. I'd suggest to merge this as a first step.

oourfali:
@jhernand I'll merge once ci finishes.

jhernand:
Please don't merge, checking the connection works, but using it doesn't :-( .

agrare:
@jhernand confirmed this works for validation but not for normal connections.  Looks like just fallout from moving a number of methods to class methods, let me know if there is anything I can do to help.

jhernand:
@agrare would be nice if you can take a look and help find the right Ruby way to call the `resolve_ip_address` method from both instance and class methods. The typical `self.resolve_ip_address` doesn't work. I guess that is related to the fact that the methods are defined in a concern, and using the `class_methods` mechanism. I am trying now to call them with `self.class.resolve_ip_address`, but I have to admit that I am a bit lost.

agrare:
@jhernand you're right about `self.class.resolve_ip_address` but it wasn't working because they're private class methods.

Try with this patch:

```
diff --git a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
index d917e43..5c7d6e0 100644
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -41,8 +41,8 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
     # Starting with version 4 of oVirt authentication doesn't work when using directly the IP address, it requires
     # the fully qualified host name, so if we received an IP address we try to convert it into the corresponding
     # host name:
-    if resolve_ip_addresses?
-      resolved = resolve_ip_address(connect_options[:server])
+    if self.class.resolve_ip_addresses?
+      resolved = self.class.resolve_ip_address(connect_options[:server])
       if resolved != connect_options[:server]
         _log.info("IP address '#{connect_options[:server]}' has been resolved to host name '#{resolved}'.")
         default_endpoint.hostname = resolved
@@ -266,8 +266,8 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       # the fully qualified host name, so if we received an IP address we try to convert it into the corresponding
       # host name:
       resolved = server
-      if resolve_ip_addresses?
-        resolved = resolve_ip_address(server)
+      if self.class.resolve_ip_addresses?
+        resolved = self.class.resolve_ip_address(server)
         if resolved != server
           _log.info("IP address '#{server}' has been resolved to host name '#{resolved}'.")
         end
@@ -391,8 +391,6 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       href && href.split("/").last
     end
 
-    private
-
     #
     # Checks if IP address to host name resolving is enabled.
     #
```

jhernand:
@agrare `self.class` seems to work, even if I don't remove the `private`, and I really would like to preserve the `private`. Unit tests are still failing, but I am getting close. Thanks!

agrare:
@jhernand I had an issue with them being private when calling `#connect`

```
>> ExtManagementSystem.first.connect
NoMethodError: private method `resolve_ip_addresses?' called for #<Class:0x0055ca75d2e5c0>
	from /var/lib/gems/2.3.0/gems/activerecord-5.0.6/lib/active_record/dynamic_matchers.rb:21:in `method_missing'
	from /home/agrare/workspace/redhat/manageiq/plugins/manageiq-providers-ovirt/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb:44:in `connect'
```

Is this logic that we could move into the `raw_connect_v4` method?  The comment states it starts in version 4.

jhernand:
@agrare the name resolution logic can't be moved to the `raw_connect_v4` method. The reason is that the name resolution is needed in order to work with version 4 of the the oVirt server, regardless of what version of the API is used. Version 4 of the server supports versions 3 and 4 of the API, version 3 of the server only supports version 3 of the API. But the name resolution has to be done before knowing the version of the server.

jhernand:
@agrare you were right about the need to remove the `private`, unfortunately. I didn't find other way to make this work than removing it and using `self.class` to call those methods. It is ugly, but works. I marked them with `@api private`, hopefully future readers will understand.

jhernand:
Dear reviewers, this is now, to the best of my understanding, working with oVirt 3 and oVirt 4, and it passes the tests. Please re-review.

simaishi:

```
$ git log -1
commit 3bc6ba4b1f11bf7c8574e9a378dec8cbbde362ea
Date:   Tue Nov 7 09:32:13 2017 -0500

    Merge pull request #132 from jhernand/try_both_api_versions_in_raw_connect
    
    Try both API versions in `raw_connect`
    (cherry picked from commit cb185fc11567f66ad71d2287e0bc99e9507420a5)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1510504
```

PR #133, opened 2017-11-06
Add admin_ui feature support to InfraManager

vojtechszocs:
This PR is part of the following series:
- [Add ems_infra_admin_ui feature](https://github.com/ManageIQ/manageiq/pull/16403) (manageiq)
- [Add ems_infra_admin_ui feature support](https://github.com/ManageIQ/manageiq-ui-classic/pull/2644) (manageiq-ui-classic)
- **[Add admin_ui feature support to InfraManager](https://github.com/ManageIQ/manageiq-providers-ovirt/pull/133) (manageiq-providers-ovirt)**

Utilize `admin_ui` feature in `InfraManager`, using `AdminUI` concern to contain the URL generation logic.

URL to access oVirt WebAdmin is generated in two steps:
1. obtain new SSO token from oVirt Engine, using the credentials of the provider
2. construct URL to access WebAdmin, using the SSO token obtained in the previous step

This PR currently relies on Ravi's [Engine patch](https://gerrit.ovirt.org/#/c/82997/) that simplifies WebAdmin access URL construction.

Inspired by @jhernand (see original PR [here](https://github.com/ManageIQ/manageiq-providers-ovirt/pull/112)).

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
@@ -1,0 +1,101 @@
+module ManageIQ::Providers::Redhat::InfraManager::AdminUI
+  extend ActiveSupport::Concern
+
+  require 'openssl'
+  require 'ovirtsdk4'
+  require 'uri'
+  require 'json'
+
+  def queue_generate_admin_ui_url
+    task_opts = {
+      :action => 'Generate oVirt Admin UI URL for EMS'
+    }
+
+    queue_opts = {
+      :class_name  => self.class.name,
+      :instance_id => id,
+      :method_name => 'generate_admin_ui_url',
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => my_zone,
+      :args        => []
+    }
+
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def generate_admin_ui_url
+    $rhevm_log.info("Generating oVirt Admin UI URL for EMS with identifier '#{id}'.")
+
+    # Get the credentials of the provider:
+    user = authentication_userid(:default)
+    password = authentication_password(:default)
+
+    # Get the timeouts from the configuration:
+    read_timeout, open_timeout = ems_timeouts(:ems_redhat, options[:service])
+
+    # Create the HTTP client:
+    insecure = default_endpoint.verify_ssl == OpenSSL::SSL::VERIFY_NONE,
+    ca_certs = default_endpoint.certificate_authority
+    ca_certs = nil if ca_certs.blank?
+    ca_certs = [ca_certs] if ca_certs
+    client = OvirtSDK4::HttpClient.new(
+      :insecure        => insecure,
+      :ca_certs        => ca_certs,
+      :log             => $rhevm_log,
+      :timeout         => read_timeout,
+      :connect_timeout => open_timeout
+    )
+
+    # Request new SSO token using the provider credentials:
+    request = OvirtSDK4::HttpRequest.new
+    request.method = :GET
+    request.url = URI::HTTPS.build(
+      :host  => default_endpoint.hostname,
+      :port  => default_endpoint.port,
+      :path  => '/ovirt-engine/sso/oauth/token',
+      :query => URI.encode_www_form(
+        :grant_type => 'password',
+        :username   => user,
+        :password   => password,
+        :scope      => 'ovirt-app-admin ovirt-app-portal'
+      )
+    ).to_s
+    request.headers = {
+      'Accept'     => 'application/json',
+      'User-Agent' => "manageiq-providers-ovirt/#{ManageIQ::Providers::Ovirt::VERSION}"
+    }
+    client.send(request)
+    response = client.wait(request)
+
+    # Close the HTTP client:
+    client.close
+
+    # Extract the SSO token, return nil in case of error:
+    if response.is_a?(OvirtSDK4::Error) || response.code != 200
+      $rhevm_log.warn("Failed to obtain SSO token from oVirt Engine.")
+      $rhevm_log.warn(response.message) unless response.is_a?(OvirtSDK4::Error)
+      return nil
+    end
+    sso_token = JSON.parse(response.body)['access_token']
+
+    # Generate URL to access oVirt Admin UI using the SSO token:
+    app_url = URI::HTTPS.build(
+      :host => default_endpoint.hostname,
+      :port => default_endpoint.port,
+      :path => '/ovirt-engine/webadmin'
+    ).to_s
+    access_url = URI::HTTPS.build(
+      :host  => default_endpoint.hostname,
+      :port  => default_endpoint.port,
+      :path  => '/ovirt-engine/webadmin/sso/login',
+      :query => URI.encode_www_form(
+        :sso_token => sso_token,
+        :app_url   => app_url
+      )
+    ).to_s
+
+    # Return the URL:
+    access_url
+  end
+end
--- a/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
@@ -32,7 +32,7 @@ module ManageIQ::Providers::Redhat::InfraManager::AdminUI
     password = authentication_password(:default)
 
     # Get the timeouts from the configuration:
-    read_timeout, open_timeout = ems_timeouts(:ems_redhat, options[:service])
+    read_timeout, open_timeout = self.class.ems_timeouts(:ems_redhat, "Service")
 
     # Create the HTTP client:
     insecure = default_endpoint.verify_ssl == OpenSSL::SSL::VERIFY_NONE,
@@ -73,9 +73,9 @@ module ManageIQ::Providers::Redhat::InfraManager::AdminUI
     client.close
 
     # Extract the SSO token, return nil in case of error:
-    if response.is_a?(OvirtSDK4::Error) || response.code != 200
+    if response.kind_of?(OvirtSDK4::Error) || response.code != 200
       $rhevm_log.warn("Failed to obtain SSO token from oVirt Engine.")
-      $rhevm_log.warn(response.message) unless response.is_a?(OvirtSDK4::Error)
+      $rhevm_log.warn(response.message) if response.kind_of?(OvirtSDK4::Error)
       return nil
     end
     sso_token = JSON.parse(response.body)['access_token']
--- a/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
@@ -49,22 +49,23 @@ module ManageIQ::Providers::Redhat::InfraManager::AdminUI
 
     # Request new SSO token using the provider credentials:
     request = OvirtSDK4::HttpRequest.new
-    request.method = :GET
+    request.method = :POST
     request.url = URI::HTTPS.build(
-      :host  => default_endpoint.hostname,
-      :port  => default_endpoint.port,
-      :path  => '/ovirt-engine/sso/oauth/token',
-      :query => URI.encode_www_form(
-        :grant_type => 'password',
-        :username   => user,
-        :password   => password,
-        :scope      => 'ovirt-app-admin ovirt-app-portal'
-      )
+      :host => default_endpoint.hostname,
+      :port => default_endpoint.port,
+      :path => '/ovirt-engine/sso/oauth/token'
     ).to_s
     request.headers = {
-      'Accept'     => 'application/json',
-      'User-Agent' => "manageiq-providers-ovirt/#{ManageIQ::Providers::Ovirt::VERSION}"
+      'Accept'       => 'application/json',
+      'Content-Type' => 'application/x-www-form-urlencoded',
+      'User-Agent'   => "manageiq-providers-ovirt/#{ManageIQ::Providers::Ovirt::VERSION}"
     }
+    request.body = URI.encode_www_form(
+      :grant_type => 'password',
+      :username   => user,
+      :password   => password,
+      :scope      => 'ovirt-app-admin ovirt-app-portal'
+    )
     client.send(request)
     response = client.wait(request)
 
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
   require_nested  :Vm
   include_concern :ApiIntegration
   include_concern :VmImport
+  include_concern :AdminUI
 
   has_many :cloud_tenants, :foreign_key => :ems_id, :dependent => :destroy
 
@@ -21,6 +22,7 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
 
   supports :provisioning
   supports :refresh_new_target
+  supports :admin_ui
 
   def ensure_managers
     return unless enabled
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+    ssoTokenUrl = URI::HTTPS.build(
+      host: endpoint.hostname,
+      port: endpoint.port,
+      path: '/ovirt-engine/sso/oauth/token',
+      query: "grant_type=password&username=#{user}&password=#{password}&scope=#{ssoTokenScope}"
+    ).to_s
+    request = OvirtSDK4::HttpRequest.new
+    request.method = :GET
+    request.url = ssoTokenUrl
+    request.headers = { 'Accept': 'application/json' }
```
jhernand: Would be nice to have a `User-Agent` header, so that looking at the web server logs of the oVirt engine it is easy to identify this as a request coming from the ManageIQ oVirt provider.

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+
+    # Create the HTTP client:
+    insecure = endpoint.verify_ssl == OpenSSL::SSL::VERIFY_NONE,
+    ca_certs = endpoint.certificate_authority
+    ca_certs = nil if ca_certs.blank?
+    ca_certs = [ca_certs] if ca_certs
+    client = OvirtSDK4::HttpClient.new(
+      insecure: insecure,
+      ca_certs: ca_certs,
+      cookies: true,
```
jhernand: I think that your use case doesn't require cookies, so you can remove this parameter.

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+      log: $rhevm_log
+    )
+
+    # Request new SSO token using provider's credentials:
+    ssoTokenScope = 'ovirt-app-admin%20ovirt-app-portal'
```
jhernand: I'd suggest to avoid using directly that %20. Instead of that you can use the `URI.encode_www_form_component` method. Actually I'd suggest to do that for all the query parameters, as some of them may contain characters that you don't expect. For example, the password could contain a `=` character. Consider something like this:

```ruby
query = {
  :grant_type => 'password',
  :username => user,
  :password => password,
  :scope => 'ovirt-app-admin ovirt-app-portal'
}
```

Then, when building the URL:

```ruby
:query => URI.encode_www_form(query)
```

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+    # Create the queue options:
+    queue_options = {
+      class_name: self.class.name,
+      instance_id: id,
+      method_name: 'generate_ui_url',
+      priority: MiqQueue::HIGH_PRIORITY,
+      role: 'ems_operations',
+      zone: my_zone,
+      args: []
+    }
```
jhernand: I think that the style checkers (Rubocop, Code climate, etc) will tell you that they don't like this modern and nice syntax for hashes. You will have the older and not so nice syntax:

```ruby
:class_name => self.class.name,
:instance_id: => id,
...
```

They will also probably insist to have the `=>` aligned.

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+      host: endpoint.hostname,
+      port: endpoint.port,
+      path: '/ovirt-engine/webadmin',
+      fragment: 'vms'
+    ).to_s
+    uiAccessUrl = URI::HTTPS.build(
+      host: endpoint.hostname,
+      port: endpoint.port,
+      path: '/ovirt-engine/webadmin/sso/login',
+      query: "sso_token=#{ssoToken}&app_url=#{uiAppUrl}"
```
jhernand: Same comment about building the query, consider using `URI.encode_www_form`.

app/models/manageiq/providers/redhat/infra_manager/vm.rb
```diff
@@ -102,4 +107,103 @@ def has_required_host?
...
+  def queue_generate_ui_url
+    # Create the task options:
+    task_options = {
+      action: "Generate oVirt UI URL for VM '#{name}'"
+    }
+
+    # Create the queue options:
+    queue_options = {
+      class_name: self.class.name,
+      instance_id: id,
```
jhernand: When I did the initial prototype of this I placed the code in the `Vm` class because the URL was (in theory) for a specific virtual machine. But if we are going to go to the virtual machines tab, not to a specific virtual machine, it may make more sense to place this in the `InfraManager` class instead (maybe as a mixin).

vojtechszocs:
@jhernand thank you for your review comments, I think I've addressed most of them in the latest commit.

vojtechszocs:
cc @oourfali

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+    ca_certs = [ca_certs] if ca_certs
+    client = OvirtSDK4::HttpClient.new(
+      :insecure => insecure,
+      :ca_certs => ca_certs,
+      :log      => $rhevm_log
+    )
+
+    # Request new SSO token using the provider credentials:
+    request = OvirtSDK4::HttpRequest.new
+    request.method = :GET
```
jhernand: If we can use `:POST` instead of `:GET` and put the parameters in the request body, then there will be no concern about having the user name and password in the logs. That may require changes in the SSO service, not sure if it supports both `POST` and `GET` currently.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+      :query => URI.encode_www_form(
+        :grant_type => "password",
+        :username   => user,
+        :password   => password,
+        :scope      => "ovirt-app-admin ovirt-app-portal"
+      )
+    ).to_s
+    request.headers = {
+      "Accept"     => "application/json",
+      "User-Agent" => "manageiq-providers-ovirt / API version #{api_version}"
```
jhernand: The typical format of the `User-Agent` header is `name/version`, without spaces or without additional text before the version number. I'd suggest `"manageiq-providers-ovirt/#{ManageIQ::Providers::Ovirt::VERSION}"`, for example.

Note that I am not sure if we ever update that `VERSION`, we should start doing it, probably.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+    # Create the HTTP client:
+    insecure = default_endpoint.verify_ssl == OpenSSL::SSL::VERIFY_NONE,
+    ca_certs = default_endpoint.certificate_authority
+    ca_certs = nil if ca_certs.blank?
+    ca_certs = [ca_certs] if ca_certs
+    client = OvirtSDK4::HttpClient.new(
+      :insecure => insecure,
+      :ca_certs => ca_certs,
+      :log      => $rhevm_log
+    )
```
jhernand: When creating the HTTP client it may be good to honour the timeouts that are configured as part of the provider, same that we do for other HTTP connections. See here, for example: https://github.com/ManageIQ/manageiq-providers-ovirt/blob/0a05dcd81cc6f66125b396a2eb5bfa33f187358e/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb#L362-L363

```ruby

```

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+      "User-Agent" => "manageiq-providers-ovirt / API version #{api_version}"
+    }
+    client.send(request)
+    response = client.wait(request)
+
+    # Close the HTTP client:
+    client.close
+
+    # Extract the SSO token, return nil in case of error:
+    if response.code != 200
```
jhernand: The `response` can also be an exception, for example if there was a timeout when talking to the server. You may want to take that into account:

```ruby
if response.is_a?(OvirtSDK4::Error) || response.code != 200
  ...
end
```

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+      $rhevm_log.warn("Failed to obtain SSO token from oVirt Engine, server response follows.")
+      $rhevm_log.warn(response.message)
+      return nil
+    end
+    ssoToken = JSON.parse(response.body)['access_token']
```
jhernand: In Ruby variables use typically snake case: `sso_token`

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    # Generate URL to access oVirt Admin UI using the SSO token:
+    appUrl = URI::HTTPS.build(
```
jhernand: `api_url` ?

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+      :host => default_endpoint.hostname,
+      :port => default_endpoint.port,
+      :path => "/ovirt-engine/webadmin"
```
jhernand: Rubocop is going to ask you to use single quotes when there is no string interpolation or special characters: '/ovirt-engine/webadmin'.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
```diff
@@ -0,0 +1,96 @@
...
+    ).to_s
+    accessUrl = URI::HTTPS.build(
+      :host  => default_endpoint.hostname,
+      :port  => default_endpoint.port,
+      :path  => "/ovirt-engine/webadmin/sso/login",
```
jhernand: Single quotes.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: OK, I'll use those timeouts when creating the HTTP client.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: I'll test this with oVirt Engine master.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: OK, I'll change the header value to `manageiq-providers-ovirt/<version>`.

vojtechszocs: > Note that I am not sure if we ever update that `VERSION`, we should start doing it, probably.

This version is also used in `manageiq-providers-ovirt.gemspec` so it represents the oVirt Provider Ruby gem version, we can use that.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: OK, I'll update the code accordingly.

I find this kind of convention (response is either HTTP response object, or an exception) a bit weird, I'd normally expect the relevant operation (like `send`) to throw an exception and the calling code could `rescue` that exception.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: `app_url` 😃

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
vojtechszocs: I'll use the single quotes as suggested, but I haven't seen any warnings from Rubocop about those so far. Anyway, what you say makes perfect sense.

app/models/manageiq/providers/redhat/infra_manager/admin_ui.rb
jhernand: The reason for this unusual interface is that the HTTP client is inherently asynchronous, and it processes multiple requests simultaneously. When you call the `send` method you are just queuing the request, for later execution. It is later, when you call `wait` that your requests (and other that may be queued) are really processed. The `wait` method returns when your request is processed, but many other request may have already been processed as well. It isn't right to throw an exception for those other requests which may have failed. Internally the HTTP client puts your request in a table, together with the result, which can be a response or an exception. When you call `wait` you are waiting for the result of a particular request, and then looking it up in that table. So it felt natural to just return that result, regardless of what it is. Could have been more natural to check if that result is an exception, and then raise it, but preferred to keep the interface closer to the internal implementation. This allows future improvements like waiting for multiple results.

vojtechszocs: Thanks for the details!

> Could have been more natural to check if that result is an exception, and then raise it, but preferred to keep the interface closer to the internal implementation.

This is what I had in mind 😃 when calling `client.wait(request)` you're passing a specific `request` object, so I'd expect this operation to throw when that particular request fails. But keeping the interface closer to internal impl. means less code (and bugs) so I understand the motivation here.

vojtechszocs:
@jhernand aside from using HTTP `POST` instead of `GET` when obtaining new SSO token, do you see any other issues? (some are listed in #139 which tracks this PR)

vojtechszocs:
It seems that HTTP POST to `/ovirt-engine/sso/oauth/token` with URL-encoded parameters contained in the request body works as well. It's also described in [ovirt-engine-api-model docs](https://github.com/oVirt/ovirt-engine-api-model/blob/master/src/main/java/002-AuthenticationAndSecurity.adoc#authentication) 😃

vojtechszocs:
I'll address the remaining code checks, but I don't understand the one reported at admin_ui.rb:39

vojtechszocs:
This should be now ready for review.

Verified locally using a recent oVirt Engine master/4.2 build.

oourfali:

simaishi:

```
$ git log -1
commit 5c61013f5ccc1c784da1fed8bc6bb18cd54b84d3
Date:   Thu Nov 16 15:20:01 2017 +0200

    Merge pull request #133 from vojtechszocs/link-ovirt-ui
    
    Add admin_ui feature support to InfraManager
    (cherry picked from commit 3e65bcbed7272a3e11b259a31d152812a07887f2)
```

himdel:
This is broken now...

```
> ManageIQ::Providers::Redhat::InfraManager.first.supports_admin_ui?

NoMethodError: undefined method `version_higher_than?' for #<ManageIQ::Providers::Redhat::InfraManager:0x000055633f3a8908>
Did you mean?  version_at_least?
	from /home/himdel/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/activemodel-5.0.6/lib/active_model/attribute_methods.rb:433:in `method_missing'
	from /home/himdel/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/bundler/gems/manageiq-providers-ovirt-b063bedb1624/app/models/manageiq/providers/redhat/infra_manager.rb:29:in `supports_admin_ui?'
```

masayag:
@himdel this supposes to be broken only on master. The PR to rename the method from `version_higher_than?` to  `version_at_least?` wasn't backported to gaprindashvili.

Rename PR: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/143
The fix for missing rename occurrence PR: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/156

The latest two PRs weren't backported (and weren't labelled), therefore gaprindashvili should to be fine.

PR #165, opened 2017-12-03
Implement template targeted refresh

masayag:
Since we can create template from MIQ side on RHV, templates should be
fully refreshed once the creation is completed.
In addition, when a template is being created on RHV side, not all of
its fields are being updated.
Therefore, there is a need to support also template target refresh
(unrelated to graph refresh in this context).

Fixes bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1517852

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb
@@ -20,6 +20,19 @@ module ManageIQ::Providers::Redhat::InfraManager::Inventory::Strategies
       end
     end
 
+    def template_targeted_refresh(target)
+      ems.with_provider_connection(VERSION_HASH) do |connection|
+        @connection = connection
+        template_id = get_uuid_from_target(target)
+        res = {}
+        res[:cluster] = collect_clusters
+        res[:datacenter] = collect_datacenters
+        res[:template] = collect_template_by_uuid(template_id)
+        res[:storage] = target.storages.empty? ? collect_storages : collect_storage(target.storages.map { |s| get_uuid_from_target(s) })
+        res
+      end
+    end
+
     def vm_targeted_refresh(target)
       ems.with_provider_connection(VERSION_HASH) do |connection|
         @connection = connection
@@ -188,6 +201,13 @@ module ManageIQ::Providers::Redhat::InfraManager::Inventory::Strategies
     def collect_vm_by_uuid(uuid)
       vm = connection.system_service.vms_service.vm_service(uuid).get
       [VmPreloadedAttributesDecorator.new(vm, connection)]
+    rescue OvirtSDK4::Error
+      []
+    end
+
+    def collect_template_by_uuid(uuid)
+      template = connection.system_service.templates_service.template_service(uuid).get
+      [TemplatePreloadedAttributesDecorator.new(template, connection)]
     rescue OvirtSDK4::Error
       []
     end
--- a/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
@@ -24,7 +24,11 @@ module ManageIQ::Providers::Redhat::InfraManager::Refresh
           when Host
             data,  = Benchmark.realtime_block(:fetch_host_data) { host_targeted_refresh(inventory, target) }
           when VmOrTemplate
-            data,  = Benchmark.realtime_block(:fetch_vm_data) { vm_targeted_refresh(inventory, target) }
+            if target.template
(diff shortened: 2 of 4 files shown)


masayag:
@borod108 We don't have support for template target refresh, and introducing it as a specific case in https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb#L26 will require supporting it in api3 as well (which I don't think is required).

How would you suggest to address it ? Continue with the current direction or consider adding support for v3 and let the selector to determine the targeted entity type ?

masayag:
@miq-bot add_labels wip, bug

masayag:
[Bug 1517852 - Target refresh of template, further to Add template from VM is not fully populated](https://trello.com/c/WJpQZnk4/253-bug-1517852-target-refresh-of-template-further-to-add-template-from-vm-is-not-fully-populated)

masayag:

masayag:
@miq-bot remove_label wip

masayag:
@borod108, @pkliczewski decided eventually to update the code also for v3.

app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb
```diff
@@ -20,6 +20,19 @@ def host_targeted_refresh(target)
       end
     end
 
+    def template_targeted_refresh(target)
+      ems.with_provider_connection(VERSION_HASH) do |connection|
+        @connection = connection
+        template_id = get_uuid_from_target(target)
+        res = {}
+        res[:cluster] = collect_clusters
```
pkliczewski: we share 3 resources between vms and templates. What do you think about:

    def vm_targeted_refresh(target)
      ems.with_provider_connection(VERSION_HASH) do |connection|
        @connection = connection
        vm_id = get_uuid_from_target(target)
        res = {}
        res[:cluster] = collect_clusters
        res[:datacenter] = collect_datacenters
        res[:storage] = target.storages.empty? ? collect_storages : collect_storage(target.storages.map { |s| get_uuid_from_target(s) })
        if target.template ? target_template(res, vm_id) : target_vm(res, vm_id)
        res
      end
    end

    def target_template(res, template_id)
        res[:template] = collect_template_by_uuid(template_id)
    end

    def target_vm(res, template_id)
        res[:template] = search_templates("vm.id=#{vm_id}")
        res[:vm] = collect_vm_by_uuid(vm_id)
    end

app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
```diff
@@ -24,7 +24,11 @@ def collect_inventory_for_targets(ems, targets)
           when Host
             data,  = Benchmark.realtime_block(:fetch_host_data) { host_targeted_refresh(inventory, target) }
           when VmOrTemplate
-            data,  = Benchmark.realtime_block(:fetch_vm_data) { vm_targeted_refresh(inventory, target) }
+            if target.template
```
pkliczewski: if we do code like in my comment above this is not needed.

app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb
borod108: While this code is shorter, I do not think its clearer. You call vm_target_refresh but deal with a template.

pkliczewski: We deal with a template in this code anyway.

borod108: sure, but then we call it: "template_targeted_refresh"

pkliczewski: Not really, please look [here](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/inventory/strategies/v4.rb#L32)

masayag: @pkliczewski you'll have to play a similar trick for https://github.com/ManageIQ/manageiq-providers-ovirt/pull/165/files#diff-ec4f927ab832832041a9d1574a4fa6a2R21 which will make it in that case less readable

pkliczewski: Correct, it is yet another copy/paste where we change 3 lines when copying 15.

simaishi:

```
$ git log -1
commit b65b8bb81eabe1dd733d168f4eaa7e22808c8322
Date:   Thu Dec 7 13:23:20 2017 +0200

    Merge pull request #165 from masayag/template_targeted_refresh
    
    Implement template targeted refresh
    (cherry picked from commit abdb51bf050f6e1ec3251e755299075d4cea4221)
```

PR #167, opened 2017-12-04
Fix Seal option of publish VM

masayag:
Due to latest rename of property name in publish vm yaml file from :seal
to :seal_template, sealing option wasn't processed properly.

In addition, the property should be send as part of the request header,
therefore should be specified in the second parameter of the
`templates_service.add(template, :seal => options[:seal])` method.

https://bugzilla.redhat.com/show_bug.cgi?id=1514895


masayag:
@miq-bot add_labels bug, gaprindashvili/yes

masayag:
[Bug 1514895 - [RHV] Template Seal option, chosen as part of Publish VM to a template dialog is not working](https://trello.com/c/EtBEYRYQ/252-bug-1514895-rhv-template-seal-option-chosen-as-part-of-publish-vm-to-a-template-dialog-is-not-working)

PR #171, opened 2017-12-06
Targeting host fails

pkliczewski:
When ever a host is targeted event parsing fail since it is not possible to
build `vm_location`. We make sure that it is not calculated when targeting a host.

Bug-Url:
https://bugzilla.redhat.com/1520513

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parsing/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parsing/strategies/v4.rb
@@ -16,7 +16,7 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParsing::Strategies
       vm_ref = template?(event.name) ? ems_ref_from_object_in_event(event.template) : ems_ref_from_object_in_event(event.vm)
 
       # Build the event hash
-      {
+      hash = {
         :event_type          => event.name,
         :source              => 'RHEVM',
         :message             => event.description,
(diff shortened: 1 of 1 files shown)


pkliczewski:
[Bug 1520513 - Move up host to maintenance or activate host in maintenance cause error in evm.log](https://trello.com/c/tfpt3qz2/266-bug-1520513-move-up-host-to-maintenance-or-activate-host-in-maintenance-cause-error-in-evmlog)

pkliczewski:
@miq-bot add_labels gaprindashvili/yes

pkliczewski:
@miq-bot assign @masayag

pkliczewski:
@borod108 please review

app/models/manageiq/providers/redhat/infra_manager/event_parsing/strategies/v4.rb
```diff
@@ -27,18 +27,23 @@ def self.event_to_hash(event, ems_id = nil)
...
         :host_ems_ref        => ems_ref_from_object_in_event(event.host),
         :ems_cluster_ems_ref => ems_ref_from_object_in_event(event.cluster),
-        :vm_location         => vm_location(event.data_center, vm_ref)
       }
+
+      vm_location(hash, event.data_center, vm_ref)
     end
 
-    def self.vm_location(dc, vm_ref)
+    def self.vm_location(hash, dc, vm_ref)
```
borod108: I think this should either have another name or we should add the "return nil if vm_ref.nil?" without the adding the hash as a parameter and clear the vm_location key from the hash in the calling function in case its value is nil.
But to me this name does not correspond to a method that returns a hash with a lot of other attributes in addition to location, and some times does not return any location at all.

pkliczewski: I will update the name. Please note that we do not want to have nil value for this key so returning nil is not an option.

simaishi:

```
$ git log -1
commit b9376c77948b1e8d5c5e39a7e76cd1c28810e927
Date:   Thu Dec 7 13:39:03 2017 +0200

    Merge pull request #171 from pkliczewski/master
    
    Targeting host fails
    (cherry picked from commit 66d9e37c7e34d0ecb91c79c132e897e9d4e90937)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1524731
```

PR #183, opened 2017-12-24
Fix ConnectionVcr recording

borod108:
The ConnectionVCR did not save multiple results for the same request,
it saved only the result of the last one which did not allow to record
several refreshes.

This fixes the bug so now several refreshes can be recorder and played
back.


borod108:
@masayag can you please review/merge?

PR #193, opened 2018-01-11
[WIP] use EMS.destroy instead of EMS.orchestrate_destroy

jameswnl:


jameswnl:
intended change was done and merged in other PR already. closing it

PR #197, opened 2018-01-18
Migrate model display names from locale/en.yml to plugin

mzazrivec:
This is a continuation of the work started in https://github.com/ManageIQ/manageiq/pull/16596 where we want to migrate display model names from locale/en.yml into particular models, where we'll use standard gettext. Main goal here is to make this one aspect of our application pluggable.

Core PR: https://github.com/ManageIQ/manageiq/pull/16836


masayag:
@mzazrivec is it possible to add display name for templates as well ?

mzazrivec:
@masayag The changes in this PR were generated automatically by a script from the content of `locale/en.yml` (in core repo). Obviously, the display name can be added for any model, but I'd rather do that in a separate pull request.

PR #205, opened 2018-02-06
Fix adding vm_location to vm event hash

borod108:
We were not adding vm_location correctly to the event hash when
the dc in the event was blank.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1537139

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parsing/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parsing/strategies/v4.rb
@@ -37,8 +37,10 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParsing::Strategies
 
       uid_ems = ManageIQ::Providers::Redhat::InfraManager.extract_ems_ref_id(vm_ref)
       location = "#{uid_ems}.ovf"
-      return location if dc.blank?
-
+      if dc.blank?
+        hash[:vm_location] = location
+        return hash
+      end
       dc_ref = ems_ref_from_object_in_event(dc)
       dc_uid = ManageIQ::Providers::Redhat::InfraManager.extract_ems_ref_id(dc_ref)
 


pkliczewski:
The change look ok but I would check whether it affects event stream [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1232). We rely on specific structure for `vm_location`.

borod108:
@pkliczewski hi, so after talking with @agrare it seems querying the probe will be problematic at this point. Do you have any other solution for this or we will leave this as it is?

borod108:
[Bug 1537139 - TypeError: no implicit conversion of Symbol into Integer on appending EmsEvent](https://trello.com/c/xhhfQMnp/300-bug-1537139-typeerror-no-implicit-conversion-of-symbol-into-integer-on-appending-emsevent)

borod108:
@agrare can you please take a look? we had a discussion about what to do with the location, if the current way we do it is acceptable.

agrare:
So the point is that the vm_location in the event will match what the refresh_parser ends up setting in the vm's record.  If we do this will it match what the vm has after the first refresh?

pkliczewski:
@agrare According how I understand the code the answer is:  no. It won't match the vm.

agrare:
@pkliczewski @borod108 would it be possible to have the refresh_parser vm_location match the event_parser vm_location and not include the datacenter in the path?  The location just needs to uniquely identify a VM, could we maybe use the storage instead of the datacenter?

pkliczewski:
@agrare We expect [this](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1230) config path. If we change it we need to make sure to update all the vms on upgrade.

agrare:
@pkliczewski I'm good with this to fix the bug but this won't connect events to VMs that come in before the VM is in the DB (which is its own bug).

JPrause:
Is this ready to be merged. This is for a blocker issue.

simaishi:

```
$ git log -1
commit f859b70ec892c757cab0ba93793b01dbbb2d2c7e
Date:   Fri Feb 9 21:00:06 2018 -0500

    Merge pull request #205 from borod108/bugs/1537139events
    
    Fix adding vm_location to vm event hash
    (cherry picked from commit 079cd9c6db150c891bfec73fac30b20be3ed071f)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1544460
```

PR #206, opened 2018-02-08
Fix updating catalog item when provider missing

borod108:
When the provider is removed updating the catalog item would fail
because we validate if provider supports setting memory limits.

We now by default decide we do not support it if the provider is was
removed. A more elaborate fix will be discussed on the core
level.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1540592

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
@@ -137,6 +137,7 @@ class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisio
     return nil if limited.nil? || limited.zero?
 
     ems = source_ems
+    return nil if ems.blank?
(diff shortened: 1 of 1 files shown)


borod108:
[[Bug 1540592] [Regression] Cannot update Catalog Item - Save button is not working](https://trello.com/c/gKqDBrAZ/302-bug-1540592-regression-cannot-update-catalog-item-save-button-is-not-working)

simaishi:

```
$ git log -1
commit bf9a9b838244d2ac0afd058acbb2870d6409c202
Date:   Thu Feb 8 13:58:37 2018 +0200

    Merge pull request #206 from borod108/bugs/1540592updatecatalogitem
    
    Fix updating catalog item when provider missing
    (cherry picked from commit 5162ab6db39c4385eaf4d2976c878d36e9c84271)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1544487
```

PR #208, opened 2018-02-13
Update client to use renamed NetworkDiscovery

jrafanie:
https://github.com/ManageIQ/manageiq/pull/16994

PR #210, opened 2018-02-15
Add missing probe test

jrafanie:
Note, Ovirt namespace exists in the client gem and also in the
ManageIQ::Providers namespace so we need to clarify which Ovirt we want,
namely, the ovirt gem's one.

We had a regression and tests didn't detect it:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/208
https://github.com/ManageIQ/manageiq/pull/16994


borod108:
Closed and reopened to rerun specs

PR #213, opened 2018-02-25
Fix graph targeted host refresh spec

borod108:
The targeted spec would take a very long time because the requests
actually failed.


borod108:
[Target host spec taking very long](https://trello.com/c/5RMULHA6/307-target-host-spec-taking-very-long)

PR #220, opened 2018-03-01
Raise error during test if response is missing

borod108:
The connection_vcr will now raise an error if it is not in recording
mode and a response is missing.

This will make tests fail faster, on a clearer error and not rewrite a
recording in cases when it has a request missing.


borod108:
@masayag please review

pkliczewski:
It seems like we need to fix broken specs.

spec/support/ovirt_sdk/connection_vcr.rb
```diff
@@ -20,6 +20,7 @@ def wait(request)
       unless @is_recording
         res = @all_req_hash[req_key].shift
         return http_response_hash_to_obj(res) if res
+        raise "connection_vcr could not find a responce for: #{req_key}"
```
masayag: there is a typo in `responce`

PR #230, opened 2018-03-27
Properly fetch snapshot current state

borod108:
We will now fetch the snapshot current state from the snapshot_type returned
form the oVirt api, the previous calculation we used for this attribute did not
always work.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552732

### DIFF
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
@@ -483,19 +483,20 @@ describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
       expect(v.snapshots.size).to eq(3)
 
       # TODO: Fix this boolean column
-      snapshot = v.snapshots.detect { |s| s.current == 1 } # TODO: Fix this boolean column
+      snapshot = v.snapshots.detect { |s| s.name = "Active VM" } # TODO: Fix this boolean column
       expect(snapshot).to have_attributes(
         :uid         => "6e3e547f-9544-42cf-842d-9104828d8511",
         :parent_uid  => "05ff445a-0bfc-44c3-90d1-a338e9095510",
         :uid_ems     => "6e3e547f-9544-42cf-842d-9104828d8511",
         :name        => "Active VM",
         :description => "Active VM",
-        :current     => 1,
+        :current     => 0,
         :total_size  => nil,
         :filename    => nil
       )
       snapshot_parent = ::Snapshot.find_by(:name => "vm1_snap")
       expect(snapshot.parent).to eq(snapshot_parent)
+      expect(snapshot_parent.current).to eq(1)
 
       expect(v.hardware).to have_attributes(
         :guest_os             => "other",
--- a/app/models/manageiq/providers/redhat/infra_manager/refresh/parse/strategies/vm_inventory.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/refresh/parse/strategies/vm_inventory.rb
@@ -260,8 +260,9 @@ module ManageIQ::Providers::Redhat::InfraManager::Refresh::Parse::Strategies
       return result if inv.nil?
 
       parent_id = nil
-      inv.each_with_index do |snapshot, idx|
-        result << snapshot_inv_to_snapshot_hashes(snapshot, idx == inv.length - 1, parent_id)
+      inv.each do |snapshot|
+        current = snapshot.snapshot_type == "active"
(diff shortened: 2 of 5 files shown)


borod108:
@masayag @pkliczewski can you please review?
Please do not merge yet, waiting for validation.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
```diff
@@ -36,7 +36,7 @@ def load_response_mock_for(filename)
 
   it "will perform a full refresh on v4.1" do
     EmsRefresh.refresh(@ems)
-    VCR.use_cassette("#{described_class.name.underscore}_ovn_provider") do
+    VCR.use_cassette("#{described_class.name.underscore}_ovn_provider", :allow_unused_http_interactions => true) do
```
borod108: @AlonaKaplan for some reason it fails without the :allow_unused_http_interactions, any idea why that could be?

borod108: specifically on:
       There are unused HTTP interactions left in the cassette:
         - [get http://alkaplan1.usersys.redhat.com:9696/] => [200 "{\"versions\": [{\"status\": \"CURRENT\", \"id\": \"v2.0\", \"links\": [{\"href\": \"http://alk"]

pkliczewski: so make sure that vcr tape contains those calls. If we add it it is almost like having no test.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
```diff
@@ -483,19 +483,20 @@ def assert_specific_vm_powered_on
       expect(v.snapshots.size).to eq(3)
 
       # TODO: Fix this boolean column
```
pkliczewski: I think we do not need those todos anymore.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
borod108: The vcr tape contains them, the problem is it contains also other calls we do not need, but this cassette is shared with other tests and I do not know why the ovn would do less calls here.
Anyway I do not think this is like not having a test, this spec actually makes sure everything works.

pkliczewski: In this case seems like sharing is not the best idea. I am thinking here about future changes and how we are going to tack whether there are some unneeded calls that should not be there.

borod108:
[Bug 1552732 - RHV Snapshots: Reverting to "Active VM" throws "Cannot preview Active VM snapshot" in evm.log](https://trello.com/c/bIQ4CoNc/310-bug-1552732-rhv-snapshots-reverting-to-active-vm-throws-cannot-preview-active-vm-snapshot-in-evmlog)

borod108:
[Bug 1552732 - RHV Snapshots: Reverting to "Active VM" throws "Cannot preview Active VM snapshot" in evm.log](https://trello.com/c/bIQ4CoNc/310-bug-1552732-rhv-snapshots-reverting-to-active-vm-throws-cannot-preview-active-vm-snapshot-in-evmlog)

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_async_graph_spec.rb
AlonaKaplan: Hi Boris, there was a bug causing the ovn test to fail (ManageIQ/manageiq-providers-ovirt/pull/231). Are you sure that now the fix is merged ':allow_unused_http_interactions' is still needed here?

I don't like the approach of adding this flag. If actually there are cases where less interactions/calls are needed (which I don't think is the case here), a new cassette with the relevant interactions should be recorded.

borod108:
@masayag can you merge please?

simaishi:

```
$ git log -1
commit eace0377a0f90a1e2cb03ef1625a81b70603b5bd
Date:   Tue Apr 10 11:07:41 2018 +0300

    Merge pull request #230 from borod108/bugs/snapshot_active
    
    Properly fetch snapshot current state
    (cherry picked from commit 6f2a100280e060d477a775d10d2cf3427c2477b4)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1588038
```

PR #219, opened 2018-03-01
Make graph refresh enabled by default

borod108:
Graph refresh is now used for both full and targeted refresh.

PR #233, opened 2018-04-17
Use old refresh as default

borod108:
Use the old refresh as default since graph refresh is not optimized for
large environments yet.

This is due to: https://bugzilla.redhat.com/show_bug.cgi?id=1566468

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -55,7 +55,7 @@
     :user:
 :ems_refresh:
   :rhevm:
-    :inventory_object_refresh: true
+    :inventory_object_refresh: false
     :pipeline: 40
     :connections: 10
   :redhat_network:


mwperina:
Shouldn't this be pointed to 5.9 branch only?

mwperina:
And don't we need to revert also other changes which were done in https://github.com/ManageIQ/manageiq-providers-ovirt/pull/219 ?

pkliczewski:
@borod108 We were told that it worked for big amount of objects in aws. Do we know where is the issue?

pkliczewski:
@borod108 @mwperina I remember initial tests and it was faster. Do we know what has changed? Maybe it is a regression? I would not like to merge this one without understanding what is the cause of the slowdown.

borod108:
@mwperina no, they are good to have anyway.

borod108:
@pkliczewski yes, Ladas found this:
"looks like the issue is in the custom saving code of folders and clusters:
https://github.com/Ladas/manageiq-providers-ovirt/blob/b18dcd49fbbdf9979ecdb11571e9527e28f975c8/app/models/manageiq/providers/redhat/inventory_collection_default/infra_manager.rb#L134

https://github.com/Ladas/manageiq-providers-ovirt/blob/b18dcd49fbbdf9979ecdb11571e9527e28f975c8/app/models/manageiq/providers/redhat/inventory_collection_default/infra_manager.rb#L186"

So we know what to fix, just did not do it yet.

mwperina:
> @borod108 @mwperina I remember initial tests and it was faster. Do we know what has changed?
> Maybe it is a regression? I would not like to merge this one without understanding what is the cause 
> of the slowdown.

@pkliczewski  Here is the bug with details https://bugzilla.redhat.com/1566468 
we need to postpone the switch to next z-stream, that's why we need to revert the change today in 5.9.z branch

masayag:
@borod108 please monitor travis result on gaprindashvili to assure no specs were broken.

mwperina:
@simaishi Satoe could you please backport to 5.9.2? We need to revert switching RHV provider to graph refresh by default in 5.9.2

simaishi:

```
$ git log -1
commit 414daf610891fb327eb0cf4e70e83b29099b0de2
Date:   Tue Apr 17 14:59:07 2018 +0300

    Merge pull request #233 from borod108/revert_refresh_default
    
    Use old refresh as default
    (cherry picked from commit ad3a6496af6303f8b7186981d8dffcc5860109b7)
```

mwperina:
@simaishi Satoe could you please backport to 5.9.2? We need to revert switching RHV provider to graph refresh by default in 5.9.2

mwperina:
@simaishi Sorry, I had that opened twice and now submitted 2nd time by error

PR #235, opened 2018-04-23
Add .yamllint from provider generator

jrafanie:
Add .yamllint from provider generator

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,13 @@
+---
+ignore: |
+  /vendor/**
+  /spec/manageiq/**
+  /spec/vcr_cassettes/**
+
+extends: relaxed
+
+rules:
+  indentation:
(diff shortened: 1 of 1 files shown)


borod108:
@jrafanie can you please explain what this is used for? specifically why do we ignore vcr_cassets? we have some recordings which are not part of vcr, should we ignore them to?

jrafanie:
> @jrafanie can you please explain what this is used for? specifically why do we ignore vcr_cassets? we have some recordings which are not part of vcr, should we ignore them to?

Many repos use a linter for yaml files via the miq-bot to make sure they follow some standards.  These values are the defaults common to many other ManageIQ repos.  Each repo can customize this as they best see fit.

borod108:
@jrafanie I see, thank you. So we will modify it to our need later.

PR #239, opened 2018-05-14
Update Ovirt for the new switch modeling

agrare:
Update the Ovirt refresh_parser and graph refresh parsing/persister for the split out host-local switch model.

Depends on: https://github.com/ManageIQ/manageiq/pull/17420


app/models/manageiq/providers/redhat/infra_manager/refresh/parse/parser.rb
```diff
@@ -145,6 +145,7 @@ def self.host_inv_to_hashes(inv, ems_inv, cluster_uids, _storage_uids)
         :ems_cluster      => cluster_uids[host_inv.attributes.fetch_path(:cluster, :id)],
         :hardware         => hardware,
         :switches         => switches,
+        :host_switches    => switches,
```
pkliczewski: Would be great to be consistent with naming. We use `host_virtual_switches` everywhere except of here. What do you think about making it consistent?

PR #240, opened 2018-05-15
Improve performance of EmsFolders and EmsClusters child save blocks

agrare:
Improve the performance of the ems_folders_children and ems_clusters_children custom save blocks.

Note: link_ems_inventory is equivalent to the sum of ems_folder_children and ems_cluster_children save blocks.

Tested against a live scale environment with three configurations:
master - legacy ems_refresh
save_inventory: 250.93 sec
link_ems_inventory: 27.698 sec

master - graph refresh
save_inventory: 421.19 sec
ems_folder_children + ems_cluster_children: 286.32 sec

this pr - graph refresh
save_inventory: 196.86 sec
ems_folder_children + ems_cluster_children: 46.89 sec

The graph refresh link_ems_inventory equivalent was significantly slower than the ems_refresh link_ems_inventory, but not enough so that it drowned out the rest of the graph refresh improvements allowing graph refresh to still be faster than legacy save_inventory.


agrare:
I suspect the extra db lookups in the graph refresh custom save blocks are the culprit for the slower performance on graph refresh.  I am going to try replacing those with lazy_finds to hopefully save a db lookup if we have the id cached already.

PR #244, opened 2018-05-17
Ems metadata improvements

agrare:
Consolidate the vm/host/root folder collections into a single standard ems_folders collection and rewrite ems_clusters_children to do a single query for all clusters and save in a single transaction instead of a transaction per cluster.

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -73,7 +73,7 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
     collector.datacenters.each do |datacenter|
       ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(datacenter.href)
 
-      persister.root_folders.find_or_build('root_dc').assign_attributes(
+      persister.ems_folders.find_or_build('root_dc').assign_attributes(
         :name    => 'Datacenters',
         :type    => 'EmsFolder',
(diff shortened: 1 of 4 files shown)


pkliczewski:
Travis fails with snapshot refresh issue. @borod108 Do we have different PR to fix ti?

pkliczewski:
This PR looks good. Let's understand why travis fails before merging it.

agrare:
@pkliczewski yeah we saw that test sporadically failing on master before, not sure what's up with that yet.

agrare:
@pkliczewski @borod108 looks like the problem is there are multiple snapshots named `"Active VM"` so `.detect` by name is unreliable.

```
-      snapshot = v.snapshots.detect { |s| s.name = "Active VM" } # TODO: Fix this boolean column
+      active_snapshots = v.snapshots.select { |s| s.name = "Active VM" } # TODO: Fix this boolean column
+      expect(active_snapshots.count).to eq(1)
+
```
```
  1) ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher will perform a full refresh on v4.1
     Failure/Error: expect(active_snapshots.count).to eq(1)
     
       expected: 1
            got: 3
```

borod108:
@agrare thank you for the great patch! Did you run this with the scale spec? does it look better?

agrare:
@borod108 yes this took about a second out so not as dramatic as the first patch but still a little better.

agrare:
@pkliczewski @borod108 I'm working on another approach that should be even faster though, https://github.com/ManageIQ/manageiq-providers-vmware/pull/271
Saving this top down made a huge improvement.

PR #246, opened 2018-05-21
Fix sporadic snapshot test failure

agrare:
There was a snapshot test using .detect {} to find a snapshot by name.
The problem was there were three snapshots with that name so
sporadically the wrong snapshot would be returned.


agrare:
cc @pkliczewski @borod108

PR #249, opened 2018-05-22
[GAPRINDASHVILI] [WIP] Don't call Gem::Version.new() with a nil version

agrare:
Just testing that this fixes the G branch

If the EMS api_version is nil don't call Gem::Version.new() to prevent
the exception:
```
ArgumentError:
  Malformed version number string
```

This behavior changed in gem version 2.7.7

https://travis-ci.org/ManageIQ/manageiq-providers-ovirt/builds/380600591#L1969


agrare:
Passed on travis, closing

PR #257, opened 2018-06-05
Use Graph refresh by default

borod108:
Use graph refresh as the default.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1553383
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1550579


borod108:
@masayag please merge if its ok and tests pass.

PR #265, opened 2018-06-21
Use the same assoc for targeted and full refresh

agrare:
The full refresh persister was using the :vm_and_template_ems_custom_fields
association but the targeted persister was using the :custom_attributes
association.  This caused targeted refresh to clear all
miq_custom_attributes which were added outside of refresh.

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
+++ b/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
@@ -6,7 +6,13 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
 
     @collection_group = nil
 
-    add_collection(infra, :custom_attributes)
+    add_collection(infra, :vm_and_template_ems_custom_fields) do |builder|
+      builder.add_properties(
+        :model_class => ::CustomAttribute,
+        :manager_ref => %i(name),
+      )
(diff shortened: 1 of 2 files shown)


agrare:
Looks like this is not needed back on Gaprindashvili

slemrmartin:
@borod108 Please don't merge (in the future) before review ends..

Fix should look like https://github.com/ManageIQ/manageiq-providers-ovirt/pull/267, please review, I realized that this can lead to bad `strategy` property

agrare:
@slemrmartin @borod108 and I tested this on a live system and with the added spec test, I asked him to merge it when it passed those tests and you stated that you thought it looked good on gitter.

slemrmartin:
@agrare ok then, but look at #267 anyway, it's more clear solution. And if you want to fix it in gaprindashvili, removed arel is there too and causes spec error - so this way it can be "backported"

slemrmartin:
@borod108 sorry, I don't know you knew we talked about it with Adam on gitter

slemrmartin:
changes added to https://github.com/ManageIQ/manageiq-providers-ovirt/pull/268/files#diff-d512eba5610c33fabd8eefbdd6f811e2R84 instead. #267 can be closed without merging

agrare:
@simaishi here is the PR for gaprindashvili https://github.com/ManageIQ/manageiq-providers-ovirt/pull/269

PR #267, opened 2018-06-25
Use the same assoc for targeted and full refresh #2

slemrmartin:
Successor of: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/265

`:vm_and_template_ems_custom_fields` belongs to another `@collection_group`, it can result to different strategy in targeted refresh.

Original fix should remove `arel` property for `targeted?` and `:custom_attributes` IC, which replaces this one

Cc @agrare, @borod108


slemrmartin:
@miq-bot add_label bug
@miq-bot add_label gaprindashvili/no

PR #269, opened 2018-06-25
[GAPRINDASHVILI] fix miq custom attributes deleted

agrare:
The full refresh persister was using the :vm_and_template_ems_custom_fields
association but the targeted persister was using the :custom_attributes
association.  This caused targeted refresh to clear all
miq_custom_attributes which were added outside of refresh.

Master PR: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/265

https://bugzilla.redhat.com/show_bug.cgi?id=1594833

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
+++ b/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
@@ -117,8 +117,6 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
     )
     add_inventory_collection(
       infra.vm_and_template_ems_custom_fields(
-        :arel     => CustomAttribute.joins("INNER JOIN vms ON vms.id = custom_attributes.resource_id")
-         .where(:vms => { :ems_ref => manager_refs }),
         :strategy => :local_db_find_missing_references
       )
     )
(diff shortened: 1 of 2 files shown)


agrare:
@borod108 @slemrmartin please review

slemrmartin:
:+1: LGTM

app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
```diff
@@ -117,8 +117,6 @@ def add_vms_inventory_collections(manager_refs)
     )
     add_inventory_collection(
       infra.vm_and_template_ems_custom_fields(
-        :arel     => CustomAttribute.joins("INNER JOIN vms ON vms.id = custom_attributes.resource_id")
```
Ladas: hm, but without this, the targeted refresh will now delete all ```vm_and_template_ems_custom_fields```

so we should rather fix the query?

agrare: Oh it isn't automatically scoped to the target?

Ladas: no, rhev is not using the new targeted refresh, so it's all defined manually here

simaishi:
Follow up PR: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/271

PR #270, opened 2018-07-02
Test targeted refresh clearing other custom attrs

agrare:
Add a test to catch if a targeted refresh deletes another VM's custom
attributes.


spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_4_custom_attributes_spec.rb
```diff
@@ -67,4 +67,22 @@ def load_response_mock_for(filename)
...
+
+    vm = Vm.where(:name => 'my-cirros-vm').first
+    other_vm = Vm.find_by(:name => 'vm2323')
+
+    expect(other_vm.ems_custom_attributes.count).to eq(1)
+
+    @ems.reload
+    EmsRefresh.refresh(vm)
+
+    expect(other_vm.ems_custom_attributes.count).to eq(1)
```
Ladas: can you try ```other_vm.reload.ems_custom_attributes.count```

agrare: Done, still passes

PR #271, opened 2018-07-02
[GAPRINDASHVILI] Fix targeted refresh clearing other VMs ems_custom_attrs

agrare:
Fix an issue where targeted refresh is clearing other VM's
ems_custom_attributes

Test taken from https://github.com/ManageIQ/manageiq-providers-ovirt/pull/270


agrare:
@Ladas :+1: this passes your additional spec

PR #273, opened 2018-07-04
Switch to new targeted refresh format

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/17667
- [x] https://github.com/ManageIQ/manageiq-providers-ovirt/pull/272

Switch to new targeted refresh format. Using :manager_uuids and :targeted_arel where needed. The rest is auto-generated. This replaces the manual :arel definitions everywhere.


Ladas:
cc @slemrmartin

app/models/manageiq/providers/redhat/inventory/persister.rb
```diff
@@ -22,6 +22,10 @@ def strategy
     nil
   end
 
+  def targeted?
```
slemrmartin: Not necessary, defined in base class

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/cluster_collections.rb
```diff
@@ -1,19 +1,16 @@
...
-      end
-    end
-  end
-
   # group :ems_clusters
   def add_resource_pools
     add_collection(infra, :resource_pools) do |builder|
-      builder.add_properties(:arel => manager.resource_pools.where(:uid_ems => manager_refs.collect { |ref| "#{URI(ref).path.split('/').last}_respool" })) if targeted?
+      if targeted?
+        builder.add_targeted_arel(
```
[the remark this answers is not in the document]
Ladas: targeted arel is used only in targeted mode. we can then remove the ```targeted?``` condition, but we will need to get rid of the references(:ems_clusters), or at least make it so it doesn't fail in full refresh

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/vms_collections.rb
```diff
@@ -17,76 +13,18 @@ def add_miq_templates
...
-
-  # group :vms
-  def add_operating_systems
-    add_collection(infra, :operating_systems) do |builder|
-      if targeted?
-        builder.add_properties(:arel => OperatingSystem.joins(:vm_or_template).where('vms' => {:ems_ref => manager_refs}))
-        # bug?
-        builder.add_properties(:strategy => nil) if manager_refs.blank?
-      end
+      builder.add_properties(:manager_uuids => references(:vms)) if targeted?
```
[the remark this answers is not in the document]
Ladas: @borod108 what was the reason not to use ```references(:miq_templates)``` here?

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/hosts_collections.rb
```diff
@@ -10,58 +10,17 @@ def add_hosts
       )
 
       if targeted?
-        builder.add_properties(:arel => manager.hosts.where(:ems_ref => manager_refs)) if manager_refs.present?
+        builder.add_targeted_arel(
+          lambda do |_inventory_collection|
+            manager.hosts.where(:ems_ref => references(:hosts)) # TODO why are we using ems_ref instead of uid_ems?
```
Ladas: @borod108  @agrare what is the reason of using :ems_ref here? The manager_ref for :hosts is uid_ems, should we use :ems_ref instead?

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/networks_collections.rb
```diff
@@ -5,9 +5,12 @@ module ManageIQ::Providers::Redhat::Inventory::Persister::Definitions::InfraGrou
   def add_switches
     add_collection(infra, :switches) do |builder|
       if targeted?
-        arel = ::Switch.where(:uid_ems => manager_refs) if manager_refs.present?
-
-        builder.add_properties(:arel => arel) unless arel.nil?
+        # TODO are switches shared across emses? Seems like we weren't filling ems_id
```
Ladas: @agrare @borod108 so there is a bit disconnect between full and targeted. In full refresh we take switches related to hosts, in here we filter from all switches. Should we use :ems_id in Switch and make relation through that? Or should it always go through host?

agrare:
@borod108 this looks like it should be ready to test, can you give it a quick try on a live provider to make sure it isn't horribly broken :)

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/networks_collections.rb
agrare: @Ladas I think for ovirt we should always go through the host for switches since ovirt doesn't have any shared switches yet

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_group/hosts_collections.rb
agrare: Hm looks like this is hard-coded to use `:ems_ref` as the manager_ref [link](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/inventory/collector/target_collection.rb#L133-L135), should probably be switched to allow `:uid_ems` as well.  @borod108 would this always return `[]` if the manager_ref isn't `:ems_ref`?

agrare: Hm okay I take it back, looks like it is using the ems_ref:
```
(byebug) target.try(:manager_refs_by_association).try(:[], collection)
{:ems_ref=>#<Set: {"/ovirt-engine/api/hosts/11089411-53a2-4337-8613-7c1d411e8ae8"}>}
```

PR #275, opened 2018-08-05
Add sysprep support for ovirt

borod108:
Add sysprep specification support for vm provisioning from template
for the oVirt provider.

Depends on: https://github.com/ManageIQ/manageiq/pull/17636
Required for: https://github.com/ManageIQ/manageiq-ui-classic/pull/4211
Implements: https://bugzilla.redhat.com/show_bug.cgi?id=1553833

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
@@ -21,8 +21,7 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration
   end
 
   def configure_sysprep
-    content = get_option(:sysprep_upload_text)
-
+    content = sysprep_specification_selected? ? customization_template_content : get_option(:sysprep_upload_text)
     return unless content
     with_provider_destination { |d| d.update_sysprep!(content) }
 
@@ -37,9 +36,8 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Configuration
       configure_cpu(rhevm_vm)
       configure_host_affinity(rhevm_vm)
       configure_network_adapters
-
       sysprep_option = get_option(:sysprep_enabled)
-      if sysprep_option == 'file'
+      if sysprep_option == 'file' || sysprep_specification_selected?
         configure_sysprep
       elsif sysprep_option == 'fields'
         configure_cloud_init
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -49,6 +51,10 @@ def configure_container
 
   private
 
+  def sysprep_specification_selected?
+    options.dig(:sysprep_enabled,1) == "Sysprep Specification"
```
pkliczewski: > sysprep_option = get_option(:sysprep_enabled)

Does `sysprep_option` contain `fields` or "Sysprep Specification"? Can we replace `dig` with it?

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -21,8 +21,11 @@ def configure_cloud_init
   end
 
   def configure_sysprep
-    content = get_option(:sysprep_upload_text)
-
+    if options[:sysprep_enabled][1] == "Sysprep Specification"
```
pkliczewski: It looks like we could use `sysprep_specification_selected?` here.

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -49,6 +47,20 @@ def configure_container
 
   private
 
+  def sysprep_specification_selected?
+    options.dig(:sysprep_enabled, 1) == "Sysprep Specification"
+  end
+
+  def prepare_customization_template_substitution_options
+    substitution_options = super
+    substitution_options[:timezone] = extract_timezone(substitution_options[:timezone])
```
borod108: @agrare this is not nice :( perhaps you know a better way to do it? (anyway I can do it in the automation/ui side somehow defining I want only the value without the key? currently it returns ["300", "(GMT+13:00) Nuku'alofa"])

agrare: Hm I think the automate workflow should use 300 as the key and pass down the value /cc @gmcculloug ??

gmcculloug: Fields displayed as drop-downs in the UI are stored as an array with `[key, display_name]`, which is what you are seeing here.

Values like this in the options hash are normally processed in the backend using the `get_option` method which handles parsing the field and only returning the first key if it is an array.  For VMware, which is the only other provider to use a timezone drop-down today, this is handled in the backend model and the parsed value is passed to the provider.

If the only thing you need to pass for substitution is the key then I think you would do:
`substitution_options[:timezone] = get_option(substitution_options[:timezone])`

If that works for you a small refactoring of this would be:
```ruby
def prepare_customization_template_substitution_options
  super.tap do |substitution_options|
    substitution_options[:sysprep_timezone] = get_option(substitution_options[:sysprep_timezone])
  end
end
```

If you think you need to pass both key and display_name you could setup a second key here as well and there is a `get_option_last` helper method.  For example:
```ruby
substitution_options[:sysprep_timezone_name] = get_option_last(substitution_options[:sysprep_timezone])
```

gmcculloug: @borod108 Did my suggestion not work out?  I'm curious about the details.

PR #276, opened 2018-08-16
Fix disk_type attribute

borod108:
We were calculating the disk_type based on the sparse property of the
disk. This is incorrect, it should actually be calculated based on the
format property.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1598747


Ladas:
@borod108 looks like you'll need to adjust the specs

agrare:
Unrelated, but it is kind of sad how many duplicate parsers there are :(

PR #288, opened 2018-09-13
[WIP] Enable event flooding

jameswnl:
https://bugzilla.redhat.com/show_bug.cgi?id=1535186

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_catcher/runner.rb
@@ -40,4 +40,25 @@ class ManageIQ::Providers::Redhat::InfraManager::EventCatcher::Runner < ManageIQ
   def filtered?(event)
     filtered_events.include?(event.name)
   end
+
+  def event_dedup_key(event)
+    # referred to https://www.rubydoc.info/gems/ovirt-engine-sdk/4.0.0/OvirtSDK4/EventReader
+    {
+      :description    => event.description,
+      :code           => event.code,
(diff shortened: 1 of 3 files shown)


pkliczewski:
Do we have any known issues with flooding? In general we [get](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/lib/manageiq/providers/ovirt/legacy/event_monitor.rb#L26) the latest and we start [polling](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/0b4b6289fad7b8e9136e054f973919da8f4769a6/app/models/manageiq/providers/redhat/infra_manager/event_fetcher.rb#L10). As far as I understands how the [audit_log](https://github.com/oVirt/ovirt-engine/blob/ede62008318d924556bc9dfc5710d90e9519670d/packaging/dbscripts/create_tables.sql#L275) works I would not expect any duplicates (maybe with exception of user login but we ignore this already).

mwperina:
> Do we have any known issues with flooding? In general we get the latest and we start polling. As far as I understands how the audit_log works I would not expect any duplicates (maybe with exception of user login but we ignore this already).

Right, oVirt/RHV itself contains anti-flooding mechanism for the events stored in audit_log, so only relevant events are stored in database (and from it transferred to CFME).

@jameswnl Do we have any issues raised by customers using RHV about event flooding?

jameswnl:
This is a big raised by QE.

Let me confirm with them

jameswnl:
Closing as the BZ requirements have been changed and RHV already has this feature built internally.

PR #291, opened 2018-09-25
[GAPRINDASHVILI] fix targeted refresh

agrare:
We discovered an issue with oVirt targeted host refresh which led to lans being cleared.  We proceeded to discover more issues:

After a targeted refresh of a host:
1. Lans would be deleted
2. An extra cluster was created
3. Extra networks were created
4. Storages were duplicated
5. HostStorages were being duplicated and failing on duplicate key constraints

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1619431

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
+++ b/app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
@@ -219,7 +219,7 @@ class ManageIQ::Providers::Redhat::Inventory::Persister::TargetCollection < Mana
     )
     add_inventory_collection(
       infra.host_storages(
-        :arel     => manager.host_storages.where(:ems_ref => manager_refs),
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/redhat/inventory/persister/target_collection.rb
```diff
@@ -233,7 +233,7 @@ def add_hosts_inventory_collections(manager_refs)
     )
     add_inventory_collection(
       infra.host_networks(
-        :arel     => manager.networks.joins(:hardware => :host).where(
+        :arel     => manager.host_networks.joins(:hardware => :host).where(
```
Ladas: ah, good catch, I thought the joins were bad, but it's actually correct

agrare:
Hey @pkliczewski thanks for the review!  @simaishi should be the only merger to non-master branches since there is more that has to happen on the backend (bugzilla flags, approvals, etc...).

PR #304, opened 2018-10-29
Upgrade Ovirt Cluster through Ansibel

borod108:
Add method to upgrade cluster using ovirt-ansible-cluster-upgrade role.

Depends on: https://github.com/ManageIQ/manageiq/pull/18108

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ems_cluster.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ems_cluster.rb
@@ -1,0 +1,29 @@
+class ManageIQ::Providers::Redhat::InfraManager::EmsCluster < ManageIQ::Providers::InfraManager::EmsCluster
+  def upgrade(options = {})
+    role_options = {:role_name => "oVirt.cluster-upgrade"}
+    job = ManageIQ::Providers::AnsibleRoleWorkflow.create_job({}, extra_vars_for_upgrade(options), role_options)
+    job.signal(:start)
+    job.miq_task
+  end
+
+  private
+
+  def extra_vars_for_upgrade(options = {})
+    connect_options = ext_management_system.apply_connection_options_defaults(options)
+
+    url = URI::Generic.build(
+      :scheme => connect_options[:scheme],
+      :host   => connect_options[:server],
+      :port   => connect_options[:port],
+      :path   => connect_options[:path]
+    ).to_s
+
+    {
+      :engine_url      => url,
+      :engine_user     => connect_options[:username],
+      :engine_password => connect_options[:password],
+      :cluster_name    => name,
+      :hostname        => "localhost"
+    }
+  end
+end
--- a/spec/models/manageiq/providers/redhat/infra_manager/ems_cluster_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/ems_cluster_spec.rb
@@ -1,0 +1,21 @@
+describe ManageIQ::Providers::Redhat::InfraManager::EmsCluster do
+  context "#upgrade" do
+    before do
+      @ems = FactoryGirl.create(:ems_redhat_with_authentication)
+      @cluster = FactoryGirl.create(:ems_cluster_redhat, :ems_id => @ems.id)
+      my_server = double("my_server", :guid => "guid1")
+      allow(MiqServer).to receive(:my_server).and_return(my_server)
+    end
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/redhat/infra_manager/ems_cluster.rb
```diff
@@ -0,0 +1,27 @@
...
+      :port   => connect_options[:port],
+      :path   => connect_options[:path]
+    ).to_s
+
+    {
+      engine_url: url,
+      engine_user: connect_options[:username],
+      engine_password: connect_options[:password],
+      cluster_name: name,
+      hostname: "localhost"
```
pkliczewski: localhost?

app/models/manageiq/providers/redhat/infra_manager/ems_cluster.rb
```diff
@@ -0,0 +1,27 @@
...
+      :host   => connect_options[:server],
+      :port   => connect_options[:port],
+      :path   => connect_options[:path]
+    ).to_s
+
+    {
+      engine_url: url,
+      engine_user: connect_options[:username],
+      engine_password: connect_options[:password],
+      cluster_name: name,
```
pkliczewski: I do not see `name` being defined. Is it defined in super class?

masayag:
@borod108 typo in header - Ansibel

borod108:
@pkliczewski @masayag will you please review and merge?

pkliczewski:
@borod108 Please make sure that Travis is happy.

spec/models/manageiq/providers/redhat/infra_manager/ems_cluster_spec.rb
```diff
@@ -0,0 +1,21 @@
+describe ManageIQ::Providers::Redhat::InfraManager::EmsCluster do
+  context "#upgrade" do
+    before do
+      @ems = FactoryGirl.create(:ems_redhat_with_authentication)
+      @cluster = FactoryGirl.create(:ems_cluster_redhat, :ems_id => @ems.id)
+      my_server = double("my_server", :guid => "guid1")
+      allow(MiqServer).to receive(:my_server).and_return(my_server)
```
masayag: What ```receive(:my_server)``` stands for ? it doesn't seems like a method name

masayag:
@borod108 there's a typo in header `Ansibel` --> `Ansible`

borod108:
@agrare  @Ladas so I manged to make it work, without moving the ems_cluster code to the core repo. So everything seems to work fine, but because we do not really have sit for ems cluster I would like you to review and tell me if it is ok to leave it like this. (The solution was setting the class_model to ::EmsCluster)

spec/models/manageiq/providers/redhat/infra_manager/ems_cluster_spec.rb
borod108: It is a method name, it is used here to put a task into MiqQueue: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ansible_runner_workflow.rb#L72

Ladas:
@borod108 hm, and does it store the right :type into EmsCluster record?

app/models/manageiq/providers/redhat/infra_manager/ems_cluster.rb
```diff
@@ -0,0 +1,29 @@
+class ManageIQ::Providers::Redhat::InfraManager::EmsCluster < ::EmsCluster
```
Ladas: @borod108 lets try to define this as

module ManageIQ
  module Providers
    module Redhat
      class InfraManager
         class EmsCluster
      end
    end
  end
end

app/models/manageiq/providers/redhat/inventory/persister/definitions/infra_collections.rb
```diff
@@ -24,7 +24,9 @@ def initialize_infra_inventory_collections
   # --- IC groups definitions ---
 
   def add_clusters_group
-    add_collection(infra, :ems_clusters)
+    add_collection(infra, :ems_clusters) do |builder|
+      builder.add_properties(:model_class => ::EmsCluster)
```
Ladas: @borod108  and with the nested definition ^, lets try

```:model_class => ManageIQ::Providers::Redhat::InfraManager::EmsCluster```

and then verify it stores the right :type into the DB

PR #307, opened 2018-11-01
[GAPRINDASHVILI] fix refresh datastore iscsi

ahrechushkin:


simaishi:
@ahrechushkin Does this fix a BZ?

PR #308, opened 2018-11-03
[wip] Add support for CA in Ansible calls

borod108:
Support creating a file from the certificate attached to the provider
and pass its path to the Ansible role.

This is part of implementing: https://bugzilla.redhat.com/show_bug.cgi?id=1644605


borod108:
depends on: https://github.com/ManageIQ/manageiq/pull/18157

PR #313, opened 2018-11-21
Add support for CA for Ansible role

borod108:
When a CA string is passed as an extra_var create a temporary directory
and in it create a file with the string as its content.
Pass the path of the file as the engine_cafile to the Ansible role.

This is part of implementing: https://bugzilla.redhat.com/show_bug.cgi?id=1644605

### DIFF
--- a/app/models/manageiq/providers/redhat/ansible_role_workflow.rb
+++ b/app/models/manageiq/providers/redhat/ansible_role_workflow.rb
@@ -1,0 +1,27 @@
+class ManageIQ::Providers::Redhat::AnsibleRoleWorkflow < ManageIQ::Providers::AnsibleRoleWorkflow
+  def pre_role
+    handle_ca_string
+    super
+  end
+
+  def post_playback
+    FileUtils.remove_entry(context[:ansible_cert_dir]) if context[:ansible_cert_dir]
+    super
+  end
+
+  def handle_ca_string
+    ca_string = options[:extra_vars] && options[:extra_vars].delete(:ca_string)
+    return if ca_string.nil?
+    context[:ansible_cert_dir] = Dir.mktmpdir("rhv_ansible_workflow")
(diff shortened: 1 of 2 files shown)


borod108:
@agrare what do you think?

agrare:
@borod108 so would this be used by every RHV ansible role or should the subclass be more specifically named?

borod108:
Currently it seems it will be used by every ovirt ansible role.

app/models/manageiq/providers/redhat/ansible_role_workflow.rb
```diff
@@ -0,0 +1,36 @@
...
+    extra_vars[:engine_cafile] = create_cert_file(ca_string)
+  end
+
+  def create_cert_file(ca_string)
+    local_filename = File.join(context[:ansible_cert_dir], "ca_#{SecureRandom.hex}")
+    File.open(local_filename, 'w') {|f| f.write(ca_string) }
+    local_filename
+  end
+
+  class CertificateFileCreator
```
agrare: Is this needed anymore?

borod108: right :), removed.

borod108:
@agrare if it is ok, will you merge? https://github.com/ManageIQ/manageiq/pull/18229 depends on this one.

simaishi:
Hammer backport details:

```
$ git log -1
commit 8a867a2f46a58c47bb4673d921e71d284c1a7f53
Date:   Thu Nov 22 08:57:26 2018 -0500

    Merge pull request #313 from borod108/rfe/ca_str
    
    Add support for CA for Ansible role
    
    (cherry picked from commit aff6bd3a56909955b7f421c0ff1a8a746472eec5)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1644605
```

PR #314, opened 2018-11-28
[HAMMER] Fix ovn recording for hammer branch

borod108:
Rerecorded ovn refresh to fix the hammer branch


simaishi:
@borod108 Thanks for fixing Travis 😄

PR #316, opened 2018-12-04
Fix targeted host refresh

borod108:


agrare:
So the problem here is that `@hosts` is set by the collector initialize https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/inventory/collector.rb#L27 so that when it gets here https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/inventory/collector/target_collection.rb#L83 we don't set `@hosts` because it isn't nil so `||=` is skipped.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    def load_response_mock_for(filename)
+      prefix = described_class.name.underscore
+      YAML.load_file(File.join('spec', 'models', prefix, 'response_yamls', filename + '.yml'))
+    end
+
+    it 'does not change the host when target refresh after full refresh' do
+      stub_settings_merge(:ems_refresh => { :rhevm => {:inventory_object_refresh => true }})
+      EmsRefresh.refresh(@ems)
+      @ems.reload
+      host = Host.where(:name => "host1").first
```
agrare: Minor nit but this should be `@ems.hosts.find_by(:ems_ref => "whatever the ems ref is")` to prevent issues with test contamination

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(host.ems_id).to eq(@ems.id)
```
agrare: If we do `@ems.hosts....` then we don't even need to cover this

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      saved_host = host_to_comparable_hash(host)
+      saved_counted_models = COUNTED_MODELS.map { |m| [m.name, m.count] }
+      saved_switches = host.switches.map { |switch| [switch.uid_ems, switch.name] }
+      expect(host.networks.count).to eq(2)
+      EmsRefresh.refresh(host)
+      host.reload
+      expect(host.networks.count).to eq(2)
+      expect(saved_switches).to match_array(host.switches.map { |switch| [switch.uid_ems, switch.name] })
+      EmsRefresh.refresh(host)
```
agrare: Any reason to do the refresh twice here?  I don't have an issue if you want to make sure that targeted host refresh is idempotent but better to do `2.times { EmsRefresh.refresh(host) }` then all of these checks instead of some in the middle.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      host.reload
+      expect([["00000000-0000-0000-0000-000000000009", "ovirtmgmt"], ["5c42817f-03fb-460e-a3ab-a8770553aeee", "vlan123t"]]).to match_array(host.switches.map { |switch| [switch.uid_ems, switch.name] })
```
agrare: I know match array should be position independent but not sure how it works with nested arrays like this.  Maybe safer to check that `host.switches.map(&:uid_ems).match_array(["00000000-0000-0000-0000-000000000009", "5c42817f-03fb-460e-a3ab-a8770553aeee"])` then do `host.switches.detect { |switch| switch.uid_ems == "5c42817f-03fb-460e-a3ab-a8770553aeee"}` to check the names?

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(host.networks.count).to eq(2)
+      counted_models = COUNTED_MODELS.map { |m| [m.name, m.count] }
+      expect(saved_host).to eq(host_to_comparable_hash(host))
+      expect(saved_counted_models).to eq(counted_models)
+    end
+
+    def host_to_comparable_hash(host)
+      h = host.attributes
```
agrare: Should be able to do `host.attributes.except("updated_on", "state_changed_on")` instead of the deletes down there

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: well actually the hosts network changes between those two refreshes. So it checks that firsh refresh changes nothing, and the second does.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: right!

app/models/manageiq/providers/redhat/inventory/collector.rb
```diff
@@ -20,13 +20,13 @@ def initialize(manager, _target)
...
-    @hosts          = []
-    @vms            = []
-    @templates      = []
+    @ems_clusters   = nil
+    @networks       = nil
+    @storagedomains = nil
+    @datacenters    = nil
+    @hosts          = nil
+    @vms            = nil
+    @templates      = nil
```
agrare: I think we should do this going forward but since we want to backport this lets not introduce too much change

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
agrare: Oh okay so is it using two different vcrs?  In that case nevermind

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: done

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: well its our inner vcr implementation for the ovirtsdk, so its one yml with all the transactions, but they are different.

app/models/manageiq/providers/redhat/inventory/collector.rb
borod108: ok

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: hmm perhaps you are right - What if I use compoesed matchers instead?

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      saved_counted_models = COUNTED_MODELS.map { |m| [m.name, m.count] }
+      saved_switches = host.switches.map { |switch| [switch.uid_ems, switch.name] }
+      expect(host.networks.count).to eq(2)
+      EmsRefresh.refresh(host)
+      host.reload
+      expect(host.networks.count).to eq(2)
+      expect(saved_switches).to contain_exactly(*host.switches.map { |switch| a_collection_containing_exactly(switch.uid_ems, switch.name) })
+      EmsRefresh.refresh(host)
+      host.reload
+      expect(host.switches.map { |switch| [switch.uid_ems, switch.name] }).to contain_exactly(
```
borod108: @agrare I think this should take care of it.

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_host_spec.rb
borod108: done

borod108:
@agrare thank you for the comments. Does it look ok now?

simaishi:
Hammer backport details:

```
$ git log -1
commit e7cc0bbe8eed7a759007e4b174811cf5bb8cb607
Date:   Wed Dec 19 15:19:38 2018 -0500

    Merge pull request #316 from borod108/bugs/target_host
    
    Fix targeted host refresh
    
    (cherry picked from commit 4a90346bc0f7b4548b101e6d737e5bae02a363ab)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1658240
```

PR #331, opened 2019-01-05
Fix bug in vm targeted refresh

borod108:
The when refreshing a vm its relationship with the host was not properly refreshed.

This fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1627297

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -136,8 +136,7 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
 
       cluster = collector.collect_cluster_for_host(host)
       dc = collector.collect_datacenter_for_cluster(cluster)
-
-      persister_host = persister.hosts.find_or_build(host.id).assign_attributes(
+      persister_host = persister.hosts.find_or_build(ems_ref).assign_attributes(
         :type             => 'ManageIQ::Providers::Redhat::InfraManager::Host',
         :ems_ref          => ems_ref,
         :ems_ref_obj      => ems_ref,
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -359,7 +358,7 @@ def vms
         :memory_reserve   => vm_memory_reserve(vm),
         :raw_power_state  => template ? "never" : vm.status,
         :boot_time        => vm.try(:start_time),
-        :host             => persister.hosts.lazy_find(host_id),
+        :host             => host_ems_ref,
```
Ladas: the lazy_find should be still there, otherwise this wont work?

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_graph_target_vm_spec.rb
```diff
@@ -57,6 +57,28 @@ def load_response_mock_for(filename)
...
+      allow(Spec::Support::OvirtSDK::ConnectionVCR).to receive(:new).with(kind_of(Hash)) do |opts|
+        Spec::Support::OvirtSDK::ConnectionVCR.new(opts,
+                                                   'spec/vcr_cassettes/manageiq/providers/redhat/infra_manager/refresh/ovirt_sdk_refresh_graph_target_vm_with_host.yml',
+                                                   false)
+      end
+      stub_settings_merge(:ems_refresh => { :rhevm => {:inventory_object_refresh => true }})
+      EmsRefresh.refresh(@ems)
+      @ems.reload
+      vm = VmOrTemplate.where(:name => "vm1").first
+      expect(vm.ems_id).to eq(@ems.id)
```
Ladas: we should probably check if the host relation is there? The check below might be just testing the host is not there in both cases.

agrare: Looks like this is now checking that there is a host_id, I'd still like to add a check that it is actually the right host but that can be done in a followup

simaishi:
Hammer backport details:

```
$ git log -1
commit ffbbc78e4d62675ad121f3ab9b2886a0a86a6b6a
Date:   Mon Jan 7 12:51:36 2019 -0500

    Merge pull request #331 from borod108/specs/add_graph_specs
    
    Fix bug in vm targeted refresh
    
    (cherry picked from commit 6ccb7511a6214304982034930a5bb0ccb3a9f3ca)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1627297
```

PR #335, opened 2019-01-14
When cloning template base disk name on vm name

borod108:
When cloning the template base the disk name on the vm name. So the disk
name will be "<vm_name>_Disk<number_of_disk>".

Implements: https://bugzilla.redhat.com/show_bug.cgi?id=1662489


borod108:
@agrare will you take a look?

mwperina:
@miq_bot add_label hammer/yes

simaishi:
Backported to hammer via https://github.com/ManageIQ/manageiq-providers-ovirt/pull/380

PR #338, opened 2019-02-13
Fixes for host_switches

agrare:
Depends on: ~~https://github.com/ManageIQ/manageiq/pull/18427~~


agrare:
> Can you please explain a little what happened here

Right so couple of things here,

1. We're updating core refresh to save host_switches as a separate collection similar to host_storages, this is because there isn't a way to infer the switches when dealing with shared/distributed switches (https://github.com/ManageIQ/manageiq/pull/18427/files#diff-50f9feed046a3410ac5531492059a6fbR324)
2. The addition of host_id to switches (https://github.com/ManageIQ/manageiq-schema/pull/232 which was a dependency of https://github.com/ManageIQ/manageiq/pull/18427) allows us to fix graph refresh's handling of switches.  We basically never deleted switches when using graph refresh because the associations had no `:dependent => :destroy`

> Also the NetworkCollection was related to the ovn was it not?

Maybe?  Using just the `:uid_ems` to lookup the switch which is really bad and leads to linking up completely unrelated inventory (like what happened recently with clusters/resource pools).

If using the uid_ems was so that it would find switches from another EMS we'll need to revisit that, probably scoping to the network manager for the specific infra_manager.

The existing graph refresh handling of switches was known to be broken so hopefully we weren't depending on that behavior here.  /cc @Ladas

agrare:
> Also the NetworkCollection was related to the ovn was it not?

Looking into this a bit further I don't see how the `ManageIQ::Providers::Redhat::Inventory::Persister::Definitions::InfraGroup::NetworksCollections` is used by the `ManageIQ::Providers::Redhat::NetworkManager` (assuming that is what you mean by OVN).  It added the `Switch` inventory collection and the OpenStack NetworkManager doesn't save switches at all so I think we're good there.

PR #343, opened 2019-02-26
Skip VMs missing a cluster when saving relats

agrare:
Fixes the ems_cluster reference from hosts and vms on RHV 4.3 where the cluster href (and thus ems_ref) is of a different format from what can be seen from the vm/host.

The cluster.id is always consistent however which maps to the uid_ems, so we can use a secondary ref index to lazy find the cluster by the id instead of href.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1668720

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -25,7 +25,7 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
 
       ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(cluster.href)
 
-      persister.ems_clusters.find_or_build(ems_ref).assign_attributes(
+      persister.ems_clusters.build(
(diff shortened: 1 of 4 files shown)


agrare:
@Ladas @mwperina please take another look

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -25,7 +25,7 @@ def ems_clusters
 
       ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(cluster.href)
 
-      persister.ems_clusters.find_or_build(ems_ref).assign_attributes(
+      persister.ems_clusters.build(
```
agrare: Using find_or_build().assign_attributes() wasn't updating the secondary_refs because [`build_secondary_indexes_for`](https://github.com/ManageIQ/inventory_refresh/blob/v0.1.z/lib/inventory_refresh/inventory_collection/data_storage.rb#L51) was only being called on the `build` and not on the `assign_attributes`

PR #345, opened 2019-03-04
[GAPRINDASHVILI] mapping from "powering_up" to  "on"

ahrechushkin:
User story: 
"VM don't change power_state until BitLocker password not entered
but if you want enter BitLocker password, you need access to HTML5 console
and if power state unknown HTML5 console is unable
if we are mapped 'powering_up' to 'on' it's solve this problem"


agrare:
@ahrechushkin can you make this PR against the master branch, not gaprindashvili?

ahrechushkin:
> @ahrechushkin can you make this PR against the master branch, not gaprindashvili?

okay

borod108:
@ahrechushkin can you make the rubocop happy?

PR #355, opened 2019-04-04
Remote viewer console file

ahrechushkin:
Function that allows user download console file from Ovirt/RHV via SUI (if expose it in API)
I wrote it cause HTML5 console poor performance and i had often console disconnects.

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -264,7 +264,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
 
     def remote_viewer_console_file(vm)
       vm.with_provider_object(VERSION_HASH) do |vm_service|
-        #TODO: need receives protocol from UI for opens console
+        # TODO: need receives protocol from UI for opens console
         console = vm_service.graphics_consoles_service.list.first
         return Base64.encode64(vm_service.graphics_consoles_service.console_service(console.id).remote_viewer_connection_file)
       end
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -262,6 +262,14 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
       vm.with_provider_object(VERSION_HASH, &:suspend)
     end
 
+    def remote_viewer_console_file(vm)
+      vm.with_provider_object(VERSION_HASH) do |vm_service|
+        #TODO: need receives protocol from UI for opens console
+        console = vm_service.graphics_consoles_service.list.first
+        return Base64.encode64(vm_service.graphics_consoles_service.console_service(console.id).remote_viewer_connection_file)
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


agrare:
@bmclaughlin can you review?

agrare:
@ahrechushkin can you rebase your branch to not include this merge commit? `Merge pull request #3 from ahrechushkin/ahrechushkin-patch-2  …`

ahrechushkin:
@agrare yep

ahrechushkin:
on frontend i'm processing result like this and it allow me download console file with .vv extention
$window.open('data:application/x-virt-viewer;charset=utf-8;name=console;base64,' + result)

ahrechushkin:
heey

agrare:
ping @borod108

ahrechushkin:
up

agrare:
@borod108 @mwperina anybody?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -262,6 +262,14 @@ def vm_suspend(vm)
       vm.with_provider_object(VERSION_HASH, &:suspend)
     end
 
+    def remote_viewer_console_file(vm)
+      vm.with_provider_object(VERSION_HASH) do |vm_service|
+        # TODO: need receives protocol from UI for opens console
+        console = vm_service.graphics_consoles_service.list.first
+        return Base64.encode64(vm_service.graphics_consoles_service.console_service(console.id).remote_viewer_connection_file)
```
agrare: The explicit return isn't needed

app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
```diff
@@ -262,6 +262,14 @@ def vm_suspend(vm)
       vm.with_provider_object(VERSION_HASH, &:suspend)
     end
 
+    def remote_viewer_console_file(vm)
```
agrare: Since this is actually returning a string not a file maybe just `remote_viewer_console`

app/models/manageiq/providers/redhat/infra_manager/vm/remote_console.rb
```diff
@@ -16,6 +16,10 @@ def validate_remote_console_acquire_ticket(protocol, options = {})
             "#{protocol} remote console requires the vm to be running.") if options[:check_if_running] && state != "on"
     end
 
+    def remote_viewer_console_file
+      ext_management_system.ovirt_services.remote_viewer_console_file(self)
+    end
+
```
agrare: You might need to add a `remote_viewer_console_queue` method to request this over the queue like the acquire_ticket method below

agrare:
@skateman can you help out here on how this will be used from the UI side?

skateman:
@agrare I think the same way as oVirt implements this, i.e. make the user download this file, my best guess is to use something like `send_data` after the `wait_for_task` finishes.

The problem with this approach is that we can't proxy native consoles, so it's the user's responsibility to have access to the VM's host. However, this can be solved byusing purr :wink: in the future.

skateman:
I really would love to do a little bigger rework on the `Access` toolbar that could include this new button there. But I'm a little blocked by https://github.com/ManageIQ/manageiq-providers-vmware/issues/392

gekorob:
Closing 'cause superseded by https://github.com/ManageIQ/manageiq-providers-ovirt/pull/452

PR #356, opened 2019-04-04
Refresh specs refactoring, part1

slemrmartin:
This PR only removes copy pasted code

### DIFF
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresh/ovirt_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresh/ovirt_refresher_spec_common.rb
@@ -1,3 +1,5 @@
+require 'yaml'
+
 module OvirtRefresherSpecCommon
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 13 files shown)


slemrmartin:
@agrare please review ... should I request you for these reviews or somebody else?

agrare:
@slemrmartin I'm good with this but I wonder if (and stop me if this is where you're going eventually) we wouldn't need a refresher_spec_common if these were different contexts in the same refresher_spec file?

It seems like all of these
```
ls refresher_*rb
refresher_3_1_spec.rb                  refresher_graph_target_host_spec.rb      refresher_spec.rb                             refresher_target_vm_spec.rb
refresher_4_async_graph_spec.rb        refresher_graph_target_template_spec.rb  refresher_target_host_4_spec.rb
refresher_4_async_spec.rb              refresher_graph_target_vm_spec.rb        refresher_target_template_disconnect_spec.rb
refresher_4_custom_attributes_spec.rb  refresher_huge_async_spec.rb             refresher_target_vm_4_spec.rb
```

Could just be different contexts possibly with the same VCR (especially when comparing classic and graph refresh, using the same VCR and the same expects is huge)

slemrmartin:
@agrare I want to make it similar to azure/vmware specs, this is the first simplest step. 
Unfortunatelly there is some vcr customization and at least 3 endpoints accross these specs, so next step will take more time

slemrmartin:
Btw there are two custom refreshers for API v3 and v4, both with custom refresh code. 
So I think some files could be joined, some not. For example API v3 doesn't support targeted refresh.

agrare:
Okay so we could probably do something like
```
context "full refresh" do
  %w[v3 v4].each do |api_version|
    context "with #{api_version}" do
    end
  end
end

context "targeted refresh" do
  context "with v4" do
  end
end
```

?

slemrmartin:
maybe, I have to do more investigation, what is possible with current vcr data (or if I have to record new), i made this PR as a starting point and remove duplicate code first.

agrare:
Yeah I'd like to as much as possible use a single VCR and the same set of expects for the different refresh options (v3, v4, graph, non-graph, async, sync, etc...)  to prove out that they really are the same.  This is what e.g. amazon and vmware do in their specs.

PR #361, opened 2019-05-01
Simplify infra collection definitions

agrare:
All the sub groups made it hard to find where anything was, and they
weren't re-used so it wasn't reducing duplication anyway.

Eventually I'd like to get every one of these collections as close to core as possible, specifically dropping the custom arel and targeted_arel.


app/models/manageiq/providers/redhat/inventory/persister/infra_manager.rb
```diff
@@ -1,11 +1,189 @@
...
+          lambda do |_inventory_collection|
+            manager.ems_folders.where(:type => 'Datacenter').where(:ems_ref => references(:datacenters))
+          end
+        )
+      end
+    end
   end
 
-  def strategy
-    @collection_group == :datacenters ? :local_db_find_missing_references : nil
```
agrare: @borod108 where was this used?  I only ever see it being set to `nil` in the [parent persister](https://github.com/ManageIQ/manageiq-providers-ovirt/pull/361/files#diff-b3e49d765d9e4c3ce0a3b2b1afaf954aL10)

slemrmartin: @agrare it was removed in #273. `@collection_group` variable was there because of there were different strategies for some ICs. @Ladas - do you know why did you remove it?

Ladas: I don't remember :-)

PR #309, opened 2018-11-12
Fix broken network spec

borod108:
The ovn spec was broken due to openstack provider changes.
Made a new recording.

Note currently the ovn did not respond to some of the requests that had
limit=1000 as a parameter, this is a bug on the ovn side and will be
handled separately, the current recording was actually modified as if
the bug did not exist. This does not reflect on the changes introduced
in this commit but I thought it is important to note.

PR #365, opened 2019-05-15
Don't collect security groups in rhev refresh

agrare:
Wanted to get the conversation started here with @borod108 and @aufi
Is this an appropriate solution for both ovirt and openstack?

@borod108 I'd like to add a rhv version check, something like:

```
def security_groups
  manager.api_version > 4.2.6 && manager.api_version < 4.3 ? [] : super
end
```

Can you help with the correct versions to check for there?

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1649886

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/refresh/refresher.rb
@@ -10,11 +10,12 @@ module ManageIQ::Providers::Redhat::InfraManager::Refresh
       targets_with_data = targets.collect do |target|
         _log.info "Filtering inventory for #{target.class} [#{target.name}] id: [#{target.id}]..."
 
+        ems_api_version = inventory.service.version_string.match(/([\d][\.\d]+)/)
         if ems.use_graph_refresh?
           data = ManageIQ::Providers::Redhat::Inventory.build(ems, target)
 
           # TODO: remove when graph refresh supports ems updates
-          ems.api_version = inventory.service.version_string
+          ems.api_version = ems_api_version
           ems.save
         else
           case target
@@ -35,9 +36,9 @@ module ManageIQ::Providers::Redhat::InfraManager::Refresh
 
         case ems.highest_allowed_api_version
         when '3'
-          data[:ems_api_version] = {:api_version => inventory.service.version_string}
+          data[:ems_api_version] = {:api_version => ems_api_version}
         when '4'
-          data.instance_variable_set(:@ems_api_version, :api_version => inventory.service.version_string)
+          data.instance_variable_set(:@ems_api_version, :api_version => ems_api_version)
         end
 
         _log.info "Filtering inventory...Complete"
--- a/app/models/manageiq/providers/redhat/inventory/collector/network_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector/network_manager.rb
@@ -1,4 +1,8 @@
 class ManageIQ::Providers::Redhat::Inventory::Collector::NetworkManager < ManageIQ::Providers::Openstack::Inventory::Collector::NetworkManager
+  def security_groups
+    []
+  end
+
   def tenants
     @tenants = manager.openstack_handle.tenants
   end
--- a/app/models/manageiq/providers/redhat/inventory/collector/network_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector/network_manager.rb
@@ -1,9 +1,20 @@
 class ManageIQ::Providers::Redhat::Inventory::Collector::NetworkManager < ManageIQ::Providers::Openstack::Inventory::Collector::NetworkManager
   def security_groups
-    []
+    supports_security_groups? ? super : []
   end
 
   def tenants
     @tenants = manager.openstack_handle.tenants
+  end
+
+  private
+
+  def supports_security_groups?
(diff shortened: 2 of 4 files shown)


agrare:
@borod108 so the only api_version I see from that manager is `"v2"` at least in the specs, and the parent_manager.api_version is nil (again at least in the specs)

borod108:
@agrare so we need to fetch the rhv manager and do something like:
https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb#L18

The v2 is probably for neutron not for rhv.

This is an example of why I do not like this solution: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/309/files
When openstack repo updates, the provider tests do not run, and it could just be broken without anyone  knowing it :(

aufi:
From OpenStack point of view, this PR is okay (doesn't affect OSP provider). We (me, @mansam, @Ladas ) discussed that other solution could be making security groups in refresh optional for both OSP and ovirt.

agrare:
> @agrare so we need to fetch the rhv manager and do something like:
https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb#L18

@borod108 cool that's exactly what I was planning on doing, in the spec I was looking at that api_version was `nil` which is why I asked but that might just be an issue with the spec setup.

> The v2 is probably for neutron not for rhv.

And we can't tell from the neutron version if this will work or not?  It seems like it should given that the error is in neutron I think (@aufi ?)

> This is an example of why I do not like this solution: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/309/files
> When openstack repo updates, the provider tests do not run, and it could just be broken without anyone knowing it :(

This was the most minimal change I could think of making, these things will happen periodically with any inter-package dependencies and we run travis every night regardless of if a commit was made so we should know within a day of any such breakages.

If it is going multiple days/weeks without being noticed that sounds like a different issue (that no one is looking at travis for this repo).  That should be part of the normal day's workflow for any repo maintainer.

Also @aufi I'd say your team should be cognisant of the ovirt neutron dependency and manually test ovirt refresh when doing any changes to the openstack network_manager refresh as much as possible.  Obviously things will fall through the cracks but if we try to be better then we can catch probably 90% of these issues.

aufi:
> > @agrare so we need to fetch the rhv manager and do something like:
> > https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb#L18
> 
> @borod108 cool that's exactly what I was planning on doing, in the spec I was looking at that api_version was `nil` which is why I asked but that might just be an issue with the spec setup.
> 
> > The v2 is probably for neutron not for rhv.
> 
> And we can't tell from the neutron version if this will work or not? It seems like it should given that the error is in neutron I think (@aufi ?)

Yes, that's Neutron API version which is not closely related to oVirt version. I think the condition for versions based on API version is not helpful in this case.

aufi:
> > This is an example of why I do not like this solution: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/309/files
> > When openstack repo updates, the provider tests do not run, and it could just be broken without anyone knowing it :(
> 
> This was the most minimal change I could think of making, these things will happen periodically with any inter-package dependencies and we run travis every night regardless of if a commit was made so we should know within a day of any such breakages.
> 

That would be great :+1: 

> If it is going multiple days/weeks without being noticed that sounds like a different issue (that no one is looking at travis for this repo). That should be part of the normal day's workflow for any repo maintainer.
> 
> Also @aufi I'd say your team should be cognisant of the ovirt neutron dependency and manually test ovirt refresh when doing any changes to the openstack network_manager refresh as much as possible. Obviously things will fall through the cracks but if we try to be better then we can catch probably 90% of these issues.

The "dependency" is unfortunately not well defined, we noticed that the code is reused just when OpenStack changes started failing in oVirt side. We can think about Network-related issues before merging network-related changes in OpenStack provider. But I don't think we should run inventory tests against oVirt/RHV environment on OpenStack provider side. Nightly test runs of oVirt provider sounds reasonable to. It should help identify errors quickly and help us (OSP) understand what changes need to be considered as risky for oVirt.

agrare:
> The "dependency" is unfortunately not well defined, we noticed that the code is reused just when OpenStack changes started failing in oVirt provider.

Right, that sounds pretty well defined though no?  They inherit from the OpenStack NetworkManager so e.g. any change that requires you to update specs or VCRs for the NetworkManager would also impact the Ovirt provider.

agrare:
@borod108 I figured out why api_version was nil, just needed to reload `@ems` in a spec test so we should be good to go now, please take a look

simaishi:
Hammer backport details:

```
$ git log -1
commit 1505664d43400aebf055584c8ecdc061331c1144
Author: Boris Od <bodnopoz@redhat.com>
Date:   Fri May 17 05:36:04 2019 +0300

    Merge pull request #365 from agrare/bz_1649886_dont_collect_security_groups
    
    Don't collect security groups in rhev refresh
    
    (cherry picked from commit a86dcf2019dab2792c116de7c092810d0d2cf9ab)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1669023
```

PR #367, opened 2019-05-16
Unified IC definitions

slemrmartin:
Folowing #361 (left comment there)

@agrare My intention for IC definitions was to make the same structure across all repos. Then it's easier to read for anyone.  So I made this small refactoring to unify definitions with other providers.


agrare:
@slemrmartin to me it looked like needless complexity and it was actually simpler/easier to understand the other way.  I want to get as much as possible out of the per-provider persisters so a simple list of collections being added is about as simple as it gets.

agrare:
I'd actually like to get to the point of listing the collections in core so that this becomes just `Ovirt::Inventory::Persister < Providers::Inventory::Persister; end`

Re consistency with other providers, I'd like to do this for the other providers as well :)

slemrmartin:
> I'd actually like to get to the point of listing the collections in core so that this becomes just `Ovirt::Inventory::Persister < Providers::Inventory::Persister; end`
> 
> Re consistency with other providers, I'd like to do this for the other providers as well :)
@agrare 
Well, until you can define `Ovirt::Inventory::Persister < Providers::Inventory::Persister; end` in all providers, then 
1) there'll be `add_collection` for many times instead of array with one `add_collection`, more amount of the same code doesn't help with readability
2) you'll repeat the same code for `TargetedCollection` persisters, which definitions are mix of other persister's definitions.

PR #370, opened 2019-05-17
Since we validate that this is not blank, it probably shouldn't be blank

d-m-u:
We validate presence on this and all the rest of them in this file are set so this feels like it should be too. 

Found here: https://github.com/ManageIQ/manageiq/pull/18778#issuecomment-493368969,
thanks, sorry @himdel


d-m-u:
@miq-bot add_reviewer @bdunne

d-m-u:
hey @agrare ```rake db:seed``` is broken until this gets in, any chance I could get 👀 on it please?

PR #374, opened 2019-06-12
[WIP] Fix iscsi storage duplication

borod108:
Storages are identified by their location attribute, which in the case
of the iscsi was based on the id of the first lun in the volume group.
Order of the returned luns is not guarnteed so this would sometimes
change if the luns were changed on the provider side.

We will now base the location on the volume_group id.


borod108:
This pull request is almost the same as:
@ahrechushkin  https://github.com/ManageIQ/manageiq-providers-ovirt/pull/307/files
So @ahrechushkin if you want you can add the small difference (for it to work with api4 refresh) and add this bz to your pull request and I will merge it.
https://bugzilla.redhat.com/show_bug.cgi?id=1697467

borod108:
Surpassed by: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/387

PR #379, opened 2019-06-16
Add skeletal provisioning to V4

borod108:
We did not implement skeletal provisioning for V4, the idea is to create
a new VM with new disks that have the same parameters as the original
ones but are not cloned.
The VM will have the blank template as its original template.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1716929

PR #380, opened 2019-06-18
[HAMMER] When cloning template base disk name on vm name

borod108:
When cloning the template base the disk name on the vm name. So the disk
name will be "<vm_name>_Disk<number_of_disk>".

Implements: https://bugzilla.redhat.com/show_bug.cgi?id=1722073

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services/strategies/v4.rb
@@ -534,7 +534,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies
         cluster = ovirt_services.cluster_from_href(options[:cluster], connection)
         template = get
         clone = options[:clone_type] == :full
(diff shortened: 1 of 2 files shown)


borod108:
@simaishi

simaishi:
@borod108 Thanks for the PR, but what I meant was not to add the name fix to hammer, but remove the name related code from the previous PR https://github.com/ManageIQ/manageiq-providers-ovirt/pull/379. Is that possible?

https://bugzilla.redhat.com/show_bug.cgi?id=1662489 isn't approved to be included in hammer branch, so if we want to add this to hammer, it hast to be approved.

simaishi:
@borod108 created 5.10.z clone BZ https://bugzilla.redhat.com/show_bug.cgi?id=1722073. Would you mind updating the git commit msg, so the bot will comment on the correct BZ when merged?

borod108:
@simaishi, done.

simaishi:
I guess I could have just backported https://github.com/ManageIQ/manageiq-providers-ovirt/pull/335...

PR #382, opened 2019-07-02
Collecting snapshots total_size

borod108:
Implement collecting total_size from snapshots - we are not actually
collecting the "total_size" attribute since it actually does not make
sense for snapshots, rather, we are collecting the "actual_size"
attribute from the rhv api.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1608475

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/snapshot_disks_mixin.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/snapshot_disks_mixin.rb
@@ -1,0 +1,13 @@
+module ManageIQ::Providers::Redhat::InfraManager::SnapshotDisksMixin
+  def disks_total_size(connection, vm_id)
+    return nil if active_vm_snapshot?
+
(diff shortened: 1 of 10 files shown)


borod108:
@agrare or @masayag can you please review?

app/models/manageiq/providers/redhat/infra_manager/snapshot_disks_mixin.rb
```diff
@@ -0,0 +1,13 @@
...
+  def disks_total_size(connection, vm_id)
+    return nil if active_vm_snapshot?
+
+    vm_service = connection.system_service.vms_service.vm_service(vm_id)
+    snapshot_service = vm_service.snapshots_service.snapshot_service(id)
+    snapshot_service.disks_service.list.sum(&:actual_size)
+  end
+
+  def active_vm_snapshot?
+    description == "Active VM"
```
masayag: .freeze ?

PR #383, opened 2019-07-02
[WIP] Fix count changed by setup of parent ems in factory

d-m-u:
Since we are now validating that provider types exist and are supported, the ems specs needed to only use leaf subclasses (as it makes no sense to add, or test, an intermediate class) so https://github.com/ManageIQ/manageiq/pull/18842/files broke a bunch of the provider repos in which we are testing things we'll never actually use in real code.

### DIFF
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_3_1_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_3_1_spec.rb
@@ -59,7 +59,7 @@ describe ManageIQ::Providers::Redhat::InfraManager::Refresh::Refresher do
     expect(ExtManagementSystem.count).to eq(1)
     expect(EmsFolder.count).to eq(7)
     expect(EmsCluster.count).to eq(4)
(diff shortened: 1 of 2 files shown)


d-m-u:
@borod108 thanks, sorry

spec/models/manageiq/providers/redhat/infra_manager/refresh/refresher_3_1_spec.rb
```diff
@@ -59,7 +59,7 @@ def assert_table_counts
     expect(ExtManagementSystem.count).to eq(1)
     expect(EmsFolder.count).to eq(7)
     expect(EmsCluster.count).to eq(4)
-    expect(Host.count).to eq(2)
+    expect(Host.count).to eq(1)
```
agrare: @borod108 should this have changed because of a network factory?

borod108: Ah, not sure why that would happen :/

PR #385, opened 2019-07-02
Fix host ip address parsing

agrare:
There was an issue parsing the host ipaddress where if it were present
it would be overwritten by a DNS lookup, instead of doing the DNS lookup
if it were nil.  This led to a number of workarounds in the specs to
stub the Socket.getaddrinfo call which wouldn't be needed if it used the
ip addr from inventory properly.


borod108:
@agrare now it passes, what do you think?

agrare:
Just had to fix the graph refresh parser also, it was passing locally because the DNS lookup succeeded.

PR #387, opened 2019-07-15
ISCSI duplication (reopen)

ahrechushkin:
History:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/374
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/307/files

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1697467

@borod108, i can't test 'CINDER' cause i haven't it in my env :(


borod108:
@ahrechushkin nice, thank you :)

@agrare so this is the one that will cause a one time storage duplication on existing setups, should we document it somewhere?

PR #388, opened 2019-07-18
Disable all refreshes except graph refresh

borod108:
Only use graph refresh from now on.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1731037


borod108:
or @agrare  :)

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -24,8 +24,7 @@ def authentication_status_ok?(type = nil)
   end
 
   def use_graph_refresh?
-    Settings.ems_refresh.rhevm.try(:[], :inventory_object_refresh) &&
-      use_ovirt_sdk? && supported_api_versions.include?('4')
+    true
```
agrare: @borod108 can we remove this method and update all callers to just do what they would have done for the graph refresh case?

borod108: I could, I was thinking to do that together with removing tons of other code - you can see I only removed the specs to keep it safe...
Can we leave this to the big refactoring PR?

agrare: Okay

PR #389, opened 2019-07-21
Remove the option to use Version 3 of the api

borod108:
Disable the option to force use of version 3 of the RHV api.
Note it is still being used for Smart State Analysis - to be fixed soon.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1731039


borod108:
@masayag please.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -180,7 +180,7 @@ def with_version4_vm_service(vm)
   end
 
   def use_ovirt_sdk?
-    ::Settings.ems.ems_redhat.use_ovirt_engine_sdk
+    true
```
masayag: Why is this method needed if `use_ovirt_sdk?` always return true ?
Is this being overridden by any flow ?

borod108: No we do not need it at all, I will remove this at a later refactoring, together with lots of other code, see: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/388#discussion_r304888050

In this case there are other pieces of code like builders (parser builder, event builder) to be removed later as well.

PR #390, opened 2019-07-26
Fix potential bug for graph refresh

borod108:
The core checks if to use graph refresh by looking at the settings of
the specific provider, since we removed the option to disable it from
the settings we need to override that method.

Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1731037

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager.rb
@@ -28,6 +28,10 @@ class ManageIQ::Providers::Redhat::InfraManager < ManageIQ::Providers::InfraMana
     unsupported_reason_add(:vm_import, _('Cannot import to a RHV provider of version < 4.1.5')) unless version_at_least?('4.1.5')
   end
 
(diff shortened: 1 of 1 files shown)


borod108:
@agrare I did not find a specific error caused by this yet, it does seem to use the graph refresh anyway but I think it should go it anyway.

agrare:
I think it is because you are overriding the `collect_inventory_for_targets` method which is the primary place where this is handled in core, but absolutely thing we should do this still

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit b609f889eca6e9f7b801be6859e0bb4aff114823
Date:   Fri Jul 26 09:29:05 2019 -0400

    Merge pull request #390 from borod108/bugs/remove_legacy_refresh
    
    Fix potential bug for graph refresh
    
    (cherry picked from commit 79d00a0387e1ca6dd53f770b160b1a04a9545836)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1731037
```

PR #396, opened 2019-08-01
Define host as proper lazy object

Ladas:
The current definition was being nil, when doing targeted
refresh, causing the DB lookup to fail.

Fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1733513

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -339,13 +339,13 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
 
       ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(vm.href)
 
-      host = vm.try(:host) || vm.try(:placement_policy).try(:hosts).try(:first)
-      host_ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(host.href) if host.present?
+      host_obj = vm.try(:host) || vm.try(:placement_policy).try(:hosts).try(:first)
+      host_ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(host_obj.href) if host_obj.present?
 
       datacenter_id = collector.datacenter_by_cluster_id[vm.cluster.id]
       parent_folder = persister.ems_folders.lazy_find("#{datacenter_id}_vm")
       resource_pool = persister.resource_pools.lazy_find("#{vm.cluster.id}_respool") unless template
-      host          = persister.hosts.lazy_find(host_ems_ref)
+      host          = persister.hosts.lazy_find(host_ems_ref) if host_ems_ref.present?
 
       storages, disks = storages(vm)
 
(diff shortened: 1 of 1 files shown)


Ladas:
@gekorob @borod108 do we have spec now, checking that targeted refresh of Vm was refreshing the nested `lan` relation? We should have seen this fail in spec.

borod108:
@Ladas we do have tests for targeted vm refresh but no lan test there  - but we are reimplementing the networking model so we will add some tests.
Would you say this PR is ready to be merged?

Ladas:
@borod108 it's passing specs, so it should be ok? :-) I haven't tried it on real Ovirt env (I don't have any). So I am not 100% sure if it fixes the BZ :-) So best would be to write the spec, to verify it works.

agrare:
@Ladas does this end up being the same thing?

We assign host to persister_vm, pass that into vm_hardware where it gets assigned to persister_hardware.vm_or_template

Ladas:
@agrare so theoretically it should be the same, but from the log it looks like there was null instead of the lazy_object pointing to host. And given it's being fetched as `persister_hardware.vm_or_template&.host` it looks like somebody was expecting the vm_or_template to be nil.

agrare:
Hm okay, I think the hardware will fail to save without a vm_or_template_id too, maybe it would have if it hadn't failed on the switch.host already?

Ladas:
@borod108 I am trying to test this locally, but I can't get past this error on fedora 30
```
/home/.../manageiq/vendor/bundle/ruby/2.4.0/gems/activesupport-5.1.7/lib/active_support/dependencies.rb:292:in `require': libicui18n.so.58: cannot open shared object file: No such file or directory - /home/.../manageiq/vendor/bundle/ruby/2.4.0/gems/ovirt-engine-sdk-4.3.0/lib/ovirtsdk4c.so (LoadError)
```

But I do have the lib, running `locate libicui18n.so.58` shows it. So the `ovirtsdk4c.so` probably looks only under system path `/usr/lib64/`

Any idea how to get past this? Or can you just test this locally @borod108 , I can send you the rhev env this issue is appearing on.

Ladas:
@gekorob @borod108  any idea about https://github.com/ManageIQ/manageiq-providers-ovirt/pull/396#issuecomment-517596190 ? Maybe you guys can test this locally?

gekorob:
@Ladas I'll try to test locally, but currently I've got a problem on spec me too (I'm on Fedora 30 too, but think it's not related to that).

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -498,7 +499,7 @@ def vm_hardware_guest_devices(persister_hardware, vm, addresses)
       network_id = vnic_profile.dig(:network, :id) if vnic_profile
       lan = if network_id
               switch = persister.host_virtual_switches.lazy_find(
-                :host    => persister_hardware.vm_or_template&.host,
+                :host    => host,
```
Ladas: @agrare so what you're saying is that `host_ems_ref` is blank, because the host is not assigned at this point?

So the old code should be ok, we just need to check here if `persister_hardware.vm_or_template&.host` is nil and not search for the switch here if it's nil.

agrare: Yeah that's my guess at this point

Ladas: ok, updated the condition here

Ladas:
@gekorob ok I've updated the PR based on @agrare 's latest finding. Can you test it?

simaishi:
@Ladas `ivanchuk/yes`?

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit fb892ea44ad37f0035797a828e227aaad38f8db8
Date:   Thu Aug 15 08:32:53 2019 -0400

    Merge pull request #396 from Ladas/defined_host_as_proper_lazy_object
    
    Define host as proper lazy object
    
    (cherry picked from commit 29f3e8012b2f94b8066cdc5061cb6eaa1faefcb0)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1733513
```

PR #397, opened 2019-08-02
PGError does not exist anymore

gekorob:
With new versions of the gem, pg has changed the error hierarchy.
The usage of PGError was causing problems in displaying the correct
exception in case of Postgres connection errors (wrong credentials,
db versions, ...)

Related to BZ bug: https://bugzilla.redhat.com/show_bug.cgi?id=1734770

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -522,7 +522,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
         message = error.message
         message = error.message[6..-1] if message.starts_with?('FATAL:')
         message = message.strip
-        _log.warn("PG::Error: #{message}")
+        _log.warn("#{error.class.name}: #{message}")
         MiqException::MiqEVMLoginError.new(message)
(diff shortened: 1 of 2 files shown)


gekorob:
@agrare I had a problem in travis with 2.4.6 but it seems something related to bin/setup in travis and not related to the modifcation. Have you got any suggestion?

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -515,14 +515,14 @@ def resolve_ip_address(address)
...
+    # rhevm_metrics_connect_options
     def adapt_metrics_error(error)
       case error
-      when PGError
+      when PG::Error
         message = error.message
         message = error.message[6..-1] if message.starts_with?('FATAL:')
         message = message.strip
-        _log.warn("PGError: #{message}")
+        _log.warn("PG::Error: #{message}")
```
masayag: perhaps replace "PG::Error" with "#{error.class.name}" ?

gekorob: Perfect, I'll fix it

PR #399, opened 2019-08-12
Refactor runner usage due to core refactoring

Fryguy:
See https://github.com/ManageIQ/manageiq/pull/19140

Note that there was a bug here, as the method `post_playback` didn't exist in core and was thus never called (which means these temp dirs would leak)

### DIFF
--- a/app/models/manageiq/providers/redhat/ansible_role_workflow.rb
+++ b/app/models/manageiq/providers/redhat/ansible_role_workflow.rb
@@ -1,11 +1,6 @@
 class ManageIQ::Providers::Redhat::AnsibleRoleWorkflow < ManageIQ::Providers::AnsibleRoleWorkflow
-  def pre_role
+  def pre_execute
     handle_ca_string
-    super
-  end
-
-  def post_playback
-    FileUtils.remove_entry(context[:ansible_cert_dir]) if context[:ansible_cert_dir]
     super
   end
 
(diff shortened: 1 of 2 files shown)


Fryguy:
@agrare Please review.

NickLaMuro:
@Fryguy coming from https://github.com/ManageIQ/manageiq/pull/19140 for this, but should this have the same labels?  Don't want this to get left out of the `ivanchuk` release if the https://github.com/ManageIQ/manageiq/pull/19140 is required for it.

Fryguy:
I don't even know how to fix the original code, because it was basically a method that would never have been called (`post_playback`).  I guess that was support to be `post_playbook`, but I'm not sure it makes sense to fix that only to rename it right in the next method?

Fryguy:
Updated the commit message.

NickLaMuro:
"thus... thus... thus..."

:trollface: 

I think what you have there works.  Thanks.

agrare:
Yeah it was a typo, should have been `post_playbook`

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 83853a9787fb95a444866158c44aeb94f187b504
Date:   Mon Aug 12 21:39:17 2019 -0400

    Merge pull request #399 from Fryguy/refactor_runner
    
    Refactor runner usage due to core refactoring
    
    (cherry picked from commit dedfadf08f7565ca4b0a5337d25222558104fdae)
```

PR #402, opened 2019-08-28
Mock revoke_access_token

gekorob:
Hitting the real revoke_access_token implementation was causing specs to hang on some OS (depending on the tcp timeout)


borod108:
@gekorob great :) LGTM

PR #408, opened 2019-09-06
Data-Driven create params and validate_credentials

agrare:
Add a data-driven-forms params_for_create allowing dynamic API driven
UI form building and add the standard validate_credentials API.


agrare:
@borod108 we're working on allowing the API to drive what the UI forms required to add a provider requires.  Can you review this to check that I have what is required/optional and the types correct for `raw_connect`?
cc @skateman

borod108:
@agrare this is a cool change :)
We do need to have RSA key pair creds as well.

agrare:
Thanks @borod108, is this another endpoint or just another authentication type?

borod108:
@agrare just an authentication type as far as I understand.

borod108:
LGTM

PR #412, opened 2019-09-27
Allow full refresh to complete in case of async fetch failures

gekorob:
The async fetch of vm's snapshots was preventing the full refresh to
complete showing only the provder and none of the other objects.

Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1752732


borod108:
@gekorob I think in general we want to fail if the async failed since it means we did not fetch the real inventory - we discussed this at the design stage, we do not want to complete the refresh if it will result in an incomplete result on the CFME side.
If snapshots are an exception for some reason - perhaps we should have a mechanism to avoid failing for them specifically.

gekorob:
@borod108 got it. Don't know if snapshots are a particular case. I tried also to collect exceptions but decide to fail testing for a particular type of collection seemed to me a smell.

PR #415, opened 2019-10-08
Graph refresh only

gekorob:
Simplify code structure removing refresh, event parsing and inventory strategies to use only graph refresh

https://bugzilla.redhat.com/show_bug.cgi?id=1751101


borod108:
@gekorob awesome! I would suggest though - if it is possible without too much hustle, to make all the cosmetic changes in a separate PR (adding brackets etc.) because there are a lot of changes as it is, and it seems easier to follow the restructure without the small changes.
But it is really up to, the new structure looks good to me.

gekorob:
@borod108 thx for the suggestion, I'll try to separate to make things cleaner

gekorob:
@borod108 I tried to keep only code refactoring e deletions keeping lint fixes for another PR. 
Another possibility is to split refresh, event parsing and inventory refactoring into separated PRs.
Let me know. thx

app/models/manageiq/providers/redhat/infra_manager/inventory.rb
```diff
@@ -1,3 +1,133 @@
-module ManageIQ::Providers::Redhat::InfraManager::Inventory
-  require_nested :Builder
```
agrare: I don't think we can get rid of the `Inventory::Builder` it is still part of the "graph refresh api"

gekorob: I thought the require it was for the Builder tat was implementing the strategy pattern with the choice between v3 and v4.

gekorob: Hi, I've checked again and I think the "require_nested :Builder" is leading to https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/inventory/builder.rb
so according to me it was to remove due to the fact it was never called again. The collector was using directly the V4 inside strategies. 
https://github.com/ManageIQ/manageiq-providers-ovirt/blob/dfdd968d88940977db7a553a0c5b1b437007fc68/app/models/manageiq/providers/redhat/inventory/collector/infra_manager.rb#L4
hope not to be wrong.

agrare: Ah you are right, this is `infra_manager/inventory` not `inventory/`

gekorob: thx a lot @agrare for your time.

agrare:
@gekorob if this is good to go can you take out of WIP?

PR #417, opened 2019-10-22
Fix slow specs

gekorob:
Two specs were slow 'cause they were hitting low level libraries (OvirtSDK4) waiting for the timeout.
This PR introduces mock values and some lint fixes


borod108:
Nice :)

PR #423, opened 2019-11-05
Remove V3 api calls, ovirt and parallel dependencies.

gekorob:
We support now only the v4 RHV api calls.
Code has been changed removing strategies for api calls and simplifying the supported version and features.

Fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1751098


gekorob:
@miq-bot add_label wip

gekorob:
@miq-bot remove_label wip

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -163,15 +146,15 @@ def history_database_name
...
       disk_service = service.disk_attachments_service
       yield disk_service
     end
   end
 
-  def with_version4_vm_service(vm)
-    connection = connect(:version => 4)
-    service = connection.system_service.vms_service.vm_service(vm.uid_ems)
+  def with_vm_service(vm)
+    # connection = connect
```
borod108: I do not understand this comment, can you explain please?

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
```diff
@@ -29,7 +29,7 @@ def configure_sysprep
   end
 
   def configure_container
-    vm.with_provider_object(:version => vm.ext_management_system.highest_allowed_api_version) do |rhevm_vm|
+    vm.with_provider_object(:version => ManageIQ::Providers::Redhat::InfraManager::ApiIntegration::SUPPORTED_API_VERSION) do |rhevm_vm|
```
borod108: Do we still need this?
Maybe with_provider_object does not require a version at all anymore?

app/models/manageiq/providers/redhat/discovery.rb
```diff
@@ -1,19 +0,0 @@
...
-
-module ManageIQ
-  module Providers
-    module Redhat
-      class Discovery
-        def self.probe(ost)
-          ::Ovirt.logger = $rhevm_log if $rhevm_log
-
-          if ManageIQ::NetworkDiscovery::Port.open?(ost, ::Ovirt::Service::DEFAULT_PORT) &&
-             ::Ovirt::Service.ovirt?(:server => ost.ipaddr, :verify_ssl => false)
```
borod108: Sorry this is my fault, I thought you were talking about a different "Probe" class - this one we need as far as I know, it is used to "discover" providers in a range of addresses.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
gekorob: It's my fault, I commented out with the aim to remove before the commit but I forgot it. I'll fix in a while

app/models/manageiq/providers/redhat/infra_manager/provision/configuration.rb
gekorob: You're right, I removed version parameter from all the other with_provider_object and with_provider_connection but I left this one. I'll remove it

app/models/manageiq/providers/redhat/discovery.rb
gekorob: No the one we were discussing about was the probe of the OvirtSDK4. I thought that this one was not used. 
The problem with the discovery is the dependency from the "ovirt" gem. 
I think that I need to rewrite to use the OvirtSDK4.

Can I put back discovery.rb, the relative spec and the "ovirt" dependency?
I'll rewrite this part in another PR, removing also the "ovirt" gem. 
What do you think?

borod108: sounds like the right thing to do, the only thing to note is that in that case this PR does not fully close the bug, only the next on will. Which is ok.
So please add this back and I will merge, and we will close the bug with the next PR.

gekorob: Yes, I agree. Basically the old gem was checking for SSL certificates at the predefined port 443. This checks can be obtained also with plain SSL requests without using the gem.

gekorob: @borod108 I solved the problem with Discovery with the OvirtSDK4. It's quite straightforward so if you agree I can add this also to the current PR without opening a new one.

borod108: @gekorob awesome. Let me know when all the changes are in and we are ready for merge.

gekorob: @borod108 I thin we are ready. In the last commit I removed the ovirt gem and rewrote the code to use the OvirtSDK4 *exists?* that is part of the Probe class.

PR #427, opened 2019-11-10
Spec for a script that fixes storage duplication

borod108:
Depends on: https://github.com/ManageIQ/manageiq/pull/19490


borod108:
Won't fix some of the robocops in the spec since the code is just copied from an older version of the code: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/387/files#diff-46e93cadce1b824b5a392a8ee10e803f

borod108:
So we do not do specs for tools. And I thought I am so smart with the sub classing and mocking here.

PR #428, opened 2019-11-13
Add exception to the too short parameters name

gekorob:
Parameter names like 'vm' and 'dc' seem to be enough meaninful despite
they are less than three characters length


gekorob:
Proposed on the guides repo: https://github.com/ManageIQ/guides/pull/382

PR #431, opened 2019-11-14
Rely on the default implementation of .ems_class

carbonin:
Requires https://github.com/ManageIQ/manageiq/pull/19516


mwperina:
Roberto, could you please take a look?
@gekorob

Fryguy:
Merging as this is part of a big cross-repo change that goes in together with ManageIQ/manageiq#19516

PR #435, opened 2019-11-18
Remove ems_ref_obj

agrare:
The ems_ref_obj property was added for VMware because the simple ems_ref
isn't enough to fully identify an object, everyone else simply set
ems_ref_obj = ems_ref which was pointless.

https://github.com/ManageIQ/manageiq/pull/19430

### DIFF
--- a/spec/models/manageiq/providers/redhat/infra_manager/vm_import_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/vm_import_spec.rb
@@ -11,7 +11,7 @@ describe ManageIQ::Providers::Redhat::InfraManager::VmImport do
   let(:cluster_path_escaped)  { 'Folder1%2FFolder%20%40%23%24*2%2FCompute%203%2FFolder4%2FCluster%205' }
 
   let(:cluster_guid) { cluster.uid_ems }
-  let(:storage_guid) { storage.ems_ref_obj.split('/').last }
+  let(:storage_guid) { storage.ems_ref.split('/').last }
 
   let(:new_name)  { 'created-vm' }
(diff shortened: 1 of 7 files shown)


spec/models/manageiq/providers/redhat/infra_manager/vm_import_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:cluster_path_escaped)  { 'Folder1%2FFolder%20%40%23%24*2%2FCompute%203%2FFolder4%2FCluster%205' }
 
   let(:cluster_guid) { cluster.uid_ems }
-  let(:storage_guid) { storage.ems_ref_obj.split('/').last }
+  let(:storage_guid) { storage.ems_ref.split('/').last }
```
gekorob: @agrare LGTM, but storage_redhat factory is not defining ems_ref, so specs are failing.
I can see that there's an open PR in the core that removes ems_ref_obj from the factory but not replacing with ems_ref  https://github.com/ManageIQ/manageiq/pull/19430

agrare: :+1: thanks @gekorob I'll add that to the factory in the core PR and add this repo to the cross-repo tests PR

gekorob: @agrare great. So can I merge it it is it ok?

agrare: Since it depends on https://github.com/ManageIQ/manageiq/pull/19430 it'll have to be merged together

gekorob: Ok

PR #436, opened 2019-11-18
The OpenStack Network RefreshWorker was removed

agrare:
The NetworkManager::RefreshWorker was removed in https://github.com/ManageIQ/manageiq-providers-openstack/pull/538 because the CloudManager::RefreshWorker does all the work for it.  This was inheriting from the NetworkManager::RefreshWorker::Runner but ther wasn't anything overridden from the BaseManager::RefreshWorker::Runner so we can just inherit directly from that.


agrare:
cc @carbonin

agrare:
@gekorob @borod108 PTAL

gekorob:
LGTM

PR #438, opened 2019-11-18
Update specs to match behavior of change in core PR 19212 (auth user changed)

djberg96:
Depends on https://github.com/ManageIQ/manageiq/pull/19212

This breaks one current spec because the current code triggers `raise_event(:changed)` more often than it should. See https://github.com/ManageIQ/manageiq/pull/19212#issuecomment-526627820 for details.

With the PR in question, the mere act of creation would not trigger an auth changed event, only an update, so the MiqQueue numbers are lower.


gekorob:
@djberg96 LGTM

agrare:
@gekorob this had to be merged together with https://github.com/ManageIQ/manageiq/pull/19212, the CI specs on the PR are red

gekorob:
@agrare yep sorry, my fault

PR #439, opened 2019-11-18
Revert "Update specs to match behavior of change in core PR 19212 (auth user changed)"

gekorob:
Reverts ManageIQ/manageiq-providers-ovirt#438


agrare:
@gekorob even if there is an approval we don't merge our own PRs

gekorob:
@agrare ok,sorry. Understood .

PR #441, opened 2019-11-27
Script to remove storage duplication after upgrade

borod108:
We changed the way we set 'location' for block storages.
That will cause duplication of those storages during upgrade.
This script is meant to copy over the tags from the old storages and delete them.

See: ManageIQ/manageiq-providers-ovirt#387
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1697467

### DIFF
--- a/tools/handle_rhv_storage_duplication.rb
+++ b/tools/handle_rhv_storage_duplication.rb
@@ -1,0 +1,89 @@
+#!/usr/bin/env ruby
+
+# This script removes duplicate block storages that were created due to changing the way
+# storage "location" is calculated.
+# It copies the tags from the old storages and then removes them from ManageIQ.
+# See https://github.com/ManageIQ/manageiq-providers-ovirt/pull/387
+# And https://bugzilla.redhat.com/show_bug.cgi?id=1697467
+# Recommended usage: run with -dv argument to have a dry run with output.
+# If it makes sense, run with -v
+
+class HandleStorageDuplication
+  attr_reader :ems, :dry_run, :verbose
+  def initialize(opts)
+    @ems = opts[:ext_management_system]
+    @dry_run = opts[:dry_run]
+    @verbose = opts[:verbose]
+  end
+
+  def affected_storages
+    @affected_storages ||= @ems ? @ems.storages.where.not(:store_type => ["NFS", "GLANCE"]) : Storage.where.not(:store_type => ["NFS", "GLANCE"])
+  end
+
+  def handle_duplicates
+    affected_storages.each do |storage|
+      next if belongs_to_non_rhv_povider?(storage)
+
+      old_st_location = old_storage_location(storage)
+      next if old_st_location == storage.location
+
+      old_storage = Storage.where(:location => old_st_location).first
+      merge_old_storage(storage, old_storage) if old_storage
+    end
+  end
+
+  def belongs_to_non_rhv_povider?(storage)
+    storage.ext_management_system && !storage.ext_management_system.kind_of?(ManageIQ::Providers::Redhat::InfraManager)
+  end
+
+  def old_storage_location(storage)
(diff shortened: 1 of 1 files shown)


borod108:
@agrare is this ok?

borod108:
@gekorob would be happy for your review :)
Though please do not merge, since this PR is born from @agrare comments and I would like him to approve.

tools/handle_rhv_storage_duplication.rb
```diff
@@ -0,0 +1,83 @@
...
+# It copies the tags from the old storages and then removes them from ManageIQ.
+# See https://github.com/ManageIQ/manageiq-providers-ovirt/pull/387
+# And https://bugzilla.redhat.com/show_bug.cgi?id=1697467
+# Recommended usage: run with -dv argument to have a dry run with output.
+# If it makes sense, run with -v
+
+class HandleStorageDuplication
+  attr_reader :ems, :dry_run, :verbose
+  def initialize(opts)
+    @ems = opts[:ext_management_system]
```
gekorob: Only one question: Is there any other usage of the HandleStorageDuplication class? 
@ems will always be nil using HandleStorageDuplication from the script (if I understand well optimist), but maybe could be useful for testing or from other scripts.

tools/handle_rhv_storage_duplication.rb
```diff
@@ -0,0 +1,83 @@
...
+    @dry_run = opts[:dry_run]
+    @verbose = opts[:verbose]
+  end
+
+  def effected_storages
```
agrare: Should be "affected"

tools/handle_rhv_storage_duplication.rb
```diff
@@ -0,0 +1,83 @@
...
+    @effected_storages ||= @ems ? @ems.storages.where.not(:store_type => ["NFS", "GLANCE"]) : Storage.where.not(:store_type => ["NFS", "GLANCE"])
```
agrare: > Storage.where.not(:store_type => ["NFS", "GLANCE"])

Won't this also include e.g. VMware and SCVMM datastores?

tools/handle_rhv_storage_duplication.rb
borod108: oops, thanks :)

tools/handle_rhv_storage_duplication.rb
borod108: Yes it will, but they will not be found in any RHV provider so nothing will happen with them.
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/441/files#diff-1b95d354517465f92f23cd9a063b1f8dR39

Is there a better way you would suggest?

agrare: We could either take a particular provider id/name, or go through all Ovirt providers to get a list of storages in use by Ovirt

borod108: @agrare I have finally proven to myself with specs why we need to go over all of them, in Ivanchuk, before your wounderful PR for master :) the storages are not always associated with a provider, they are associated through a host - this includes the case when we get storages that are down from rhv.
In that case they are not connected to any host and I do not know how to check if they are rhv storages (perhaps use some ems_ref heuristics, but we will still need to go over all storages to do that) so we have no information on which rhv they belong to.
I have the script that runs only over connected storages ready if you will still think it is better, but I think it will case a bug in some cases.

agrare: Okay this works for me then

borod108: I added an improvement based on your comment:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/441/files#diff-1b95d354517465f92f23cd9a063b1f8dR25

If this is ok, @agrare @gekorob can you merge when specs are green?

borod108: @agrare this script makes sense only for ivanchuk, after the changes you made it makes no sense, should I add a PR to remove it after this one is back ported?

PR #444, opened 2019-12-04
Remodel to add networks by external providers

borod108:
Remodel the networking to add the vnic profiles related to networks that
are managed by external providers to the provisioning network dialog.

Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1767787
Depends on: https://github.com/ManageIQ/manageiq/pull/19576


app/models/manageiq/providers/redhat/infra_manager/external_distributed_virtual_switch.rb
```diff
@@ -0,0 +1,2 @@
+class ManageIQ::Providers::Redhat::InfraManager::ExternalDistributedVirtualSwitch < ManageIQ::Providers::InfraManager::ExternalDistributedVirtualSwitch
```
agrare: I don't think we need an external_distributed_virtual_switch at core, you can just inherit from `ManageIQ::Providers::InfraManager::DistributedVirtualSwitch` here

borod108:
closing with https://github.com/ManageIQ/manageiq-providers-ovirt/pull/450

PR #445, opened 2019-12-06
Enable to set virtio_scsi disk attachment interface

dsatsura:
Fix issue when disk_add :interface option from VmReconfigure request was not passed to DiskAttachmentBuilder

Set default disk interface to virtio_scsi according to this fix: https://bugzilla.redhat.com/show_bug.cgi?id=1399142
virtio_scsi is default interface option in oVirt/RHV UI, so make it default in ManageIQ as well.

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
@@ -12,7 +12,7 @@ class ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder
     thin_provisioned = @thin_provisioned.true?
     {
       :bootable  => @bootable.true?,
-      :interface => @interface || "virtio_scsi",
+      :interface => @interface || "VIRTIO",
       :active    => true,
       :disk      => {
         :name             => @name,
--- a/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
@@ -5,7 +5,7 @@ class ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder
     @name = options[:name]
     @thin_provisioned = BooleanParameter.new(options[:thin_provisioned])
     @bootable = BooleanParameter.new(options[:bootable])
-    @interface = options[:interface]
+    @interface = options[:interface] || ::Settings.ems.ems_ovirt.disk_attachment&.interface
   end
 
   def disk_attachment
--- a/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
@@ -5,14 +5,14 @@ class ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder
     @name = options[:name]
     @thin_provisioned = BooleanParameter.new(options[:thin_provisioned])
     @bootable = BooleanParameter.new(options[:bootable])
-    @interface = options[:interface] || ::Settings.ems.ems_ovirt.disk_attachment&.interface
+    @interface = options[:interface]
   end
 
   def disk_attachment
     thin_provisioned = @thin_provisioned.true?
     {
       :bootable  => @bootable.true?,
-      :interface => @interface || "VIRTIO",
+      :interface => @interface || "virtio_scsi",
       :active    => true,
       :disk      => {
         :name             => @name,
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
@@ -305,7 +305,11 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices
 
         # Add disks:
         added_disk_specs = spec['disksAdd']
-        add_vm_disks(vm_service, added_disk_specs) if added_disk_specs
+        if added_disk_specs
+          added_disk_specs[:default] = {}
+          added_disk_specs[:default][:interface] = 'virtio_scsi' if vm.virtio_scsi.enabled
+          add_vm_disks(vm_service, added_disk_specs)
+        end
       end
 
       _log.info("#{log_header} Completed.")
@@ -846,9 +850,10 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices
     #
     def add_vm_disks(vm_service, disk_specs)
       storage_spec = disk_specs[:storage]
+      default_disk_spec = disk_specs[:default]
       attachments_service = vm_service.disk_attachments_service
       disk_specs[:disks].each do |disk_spec|
-        attachment = prepare_vm_disk_attachment(disk_spec, storage_spec)
+        attachment = prepare_vm_disk_attachment(default_disk_spec.merge(disk_spec), storage_spec)
         attachments_service.add(attachment)
       end
     end
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
@@ -866,7 +866,8 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices
         :storage          => storage_spec,
         :name             => disk_spec[:disk_name],
         :thin_provisioned => disk_spec[:thin_provisioned],
-        :bootable         => disk_spec[:bootable]
+        :bootable         => disk_spec[:bootable],
+        :interface        => disk_spec[:interface]
       )
       attachment_builder.disk_attachment
     end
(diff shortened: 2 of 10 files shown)


app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
```diff
@@ -5,7 +5,7 @@ def initialize(options = {})
     @name = options[:name]
     @thin_provisioned = BooleanParameter.new(options[:thin_provisioned])
     @bootable = BooleanParameter.new(options[:bootable])
-    @interface = options[:interface]
+    @interface = options[:interface] || ::Settings.ems.ems_ovirt.disk_attachment&.interface
```
agrare: Not sure about a setting for this, if `"virtio-scsi"` is the default we can just do `|| "virtio-scsi"` and allow users to change it in the reconfigure UI.

agrare:
Thanks for the contribution @dsatsura !
I'll leave the final review up to the ovirt guys, we should follow this up with a change to allow users to select this in the reconfigure UI

app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb
gekorob: I think that currently the default is "VIRTIO" and set in the disk_attachment method 
https://github.com/ManageIQ/manageiq-providers-ovirt/blob/ed34bfcf8113703a09f32fc53d43c9606d4b4d85/app/models/manageiq/providers/redhat/infra_manager/disk_attachment_builder.rb#L11-L15

Maybe the default value can be changed there or remove from there a set in the constructor, as in this PR.
I agree with @agrare avoiding the setting if this value will be the default.

borod108: So as per @gekorob comment, it seems that either we remove line 15 (https://github.com/ManageIQ/manageiq-providers-ovirt/pull/445/files/64363e16bf1a1207ebc86f73a84f1b2fd717559e#L15)
or we do not touch the constructor, I am ok either way.
I agree with @agrare that we do not need a setting for this (considering we all ready do it as a string in line 15 anyway).
@dsatsura thank you very much for your contribution! - can you please describe what scenario caused an issue where the interface was not set? I think this might shed some light on the ui question.

dsatsura: Hi,
it's ok for me to set default via '|| "virtio_scsi"'.
I will update this PR and fix specs.

@borod108 disk interface always is set to **"virtio"**, no matter which **:interface** value you use in VmReconfigure request.

agrare:
:+1: LGTM

@borod108 @gekorob is there any reason virtio_scsi wouldn't be supported by all vms/hosts?  Or really any reason why we would need to stay with VIRTIO as the default...

gekorob:
@agrare I'm not a SCSI controller expert but if I'm not wrong in the ovirt doc is reported as set to true for clusters >= 3.3.
And I found that virtio-scsi is fully supported in RHEL 6.4 and later (so both host OS and guest OS must be RHEL 6.4+)
Don't know if there compatibility reasons to stay with VIRTIO for other hosts/vms OS. We should ask.

gekorob:
@dsatsura @borod108 After having talked with RHV/Ovirt, turns out that 'virtio_scsi' can be the default value but only if supported from the VM which we are re-configuring.
So I can propose to obtain the information about the virtio_scsi support and to pass it down to the DiskAttachmentBuilder in the options (something like **:virtio_scsi_enabled**)

The code of this PR so should be changed and become something like

* DiskAttachmentBuilder*
```ruby
class ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder
def initialize(options = {})
    ...
    @interface = options[:interface]
    @virtio_scsi_enabled = options[:virtio_scsi_enabled]
end

def disk_attachment
    thin_provisioned = @thin_provisioned.true?
    {
      :bootable  => @bootable.true?,
      :interface => @interface || (@virtio_scsi_enabled ? 'virtio_scsi' : 'VIRTIO'),
      ...
```
Or
```ruby
class ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder
  def initialize(options = {})
       ...
       @interface = options[:interface] || (options[:virtio_scsi_enabled] ? 'virtio_scsi' : 'VIRTIO')
  end

  def disk_attachment
    thin_provisioned = @thin_provisioned.true?
    {
      :bootable  => @bootable.true?,
      :interface => @interface,
      ...
```
All the above is valid for the default value of the interface. In case a VM have no support for 'virtio_scsi' and the interface will be set to 'virtio_scsi', the reconfiguration task will fail (cause RHV will reject the update). 

Let me know what do you think about the proposed solution.
In case we agree, I'll do a PR after this one to populate the **:virtio_scsi_enabled** option.

borod108:
@gekorob both options look great to me (I would do the second one)
How do you propose we set the virtio_scsi_enabled?

gekorob:
Hi @borod108, well in my mind came the following solution. Probably not the best and cleaner 'cause I don't like too much passing parameters in cascade.
In the *ovirt_services.rb* *vm_reconfigure* method we are already requesting details of the vm. We can add the option to get the additional contents

```ruby
    def vm_reconfigure(vm, options = {})
        ...
        vm.with_provider_object do |vm_service|
        ...
           vm = vm_service.get(:all_content => true)   # this is to get the additional values
           ...
           # Add disks
           added_disk_specs = spec['disksAdd']
           add_vm_disks(vm_service, added_disk_specs, vm.virtio_scsi.enabled) if added_disk_specs
      end
      _log.info("#{log_header} Completed.")
    end
```
The we have to propagate the value through **add_vm_disks** until **prepare_vm_disk_attachment**, because currently we have this two methods in cascade to get the disk_attachement.

```ruby
    def add_vm_disks(vm_service, disk_specs, virtio_scsi_enabled)
      storage_spec = disk_specs[:storage]
      attachments_service = vm_service.disk_attachments_service
      disk_specs[:disks].each do |disk_spec|
        attachment = prepare_vm_disk_attachment(disk_spec.merge(:virtio_scsi_enabled => virtio_scsi_enabled), 
                                                storage_spec)
        attachments_service.add(attachment)
      end
    end

    #
    # Prepares a disk attachment for adding a new disk to a virtual machine.
    #
    # @param disk_spec [Hash] The specification of the disk to add.
    # @param storage_spec [Hash] The specification of the storage to use.
    #
    def prepare_vm_disk_attachment(disk_spec, storage_spec)
      disk_spec = disk_spec.symbolize_keys
      attachment_builder = ManageIQ::Providers::Redhat::InfraManager::DiskAttachmentBuilder.new(
        :size_in_mb       => disk_spec[:disk_size_in_mb],
        :storage          => storage_spec,
        :name             => disk_spec[:disk_name],
        :thin_provisioned => disk_spec[:thin_provisioned],
        :bootable         => disk_spec[:bootable],
        :virtio_scsi_enabled => disk_spec[:virtio_scsi_enabled]
      )
      attachment_builder.disk_attachment
    end
``` 
What do you think?

borod108:
@gekorob LGTM.

gekorob:
@borod108 thanks.
So @dsatsura when you'll have time feel free to make modifications to you're PR (possibly with the second proposed solution with the assignment in the initialize method) and if you need anything just ask.
Thanks for your patience.

dsatsura:
Hi @gekorob, @agrare 
sorry for delay - was on vacation :)
what do you think about making default disk_spec with :default key in the **disk_specs** hash and leave all logic for such special defaults values inside **vm_reconfigure**.
```
    def vm_reconfigure(vm, options = {})
        ...
        vm.with_provider_object do |vm_service|
        ...
           vm = vm_service.get(:all_content => true)   # this is to get the additional values
           ...
           # Add disks
           added_disk_specs = spec['disksAdd']
           if added_disk_specs
             added_disk_specs[:default] = {}
             added_disk_specs[:default][:interface] = 'virtio_scsi' if vm.virtio_scsi.enabled
             add_vm_disks(vm_service, added_disk_specs)
           end
      end
      _log.info("#{log_header} Completed.")
    end
```
```
    def add_vm_disks(vm_service, disk_specs)
      storage_spec = disk_specs[:storage]
      default_disk_spec = disk_specs[:default]
      attachments_service = vm_service.disk_attachments_service
      disk_specs[:disks].each do |disk_spec|
        attachment = prepare_vm_disk_attachment(default_disk_spec.merge(disk_spec), 
                                                storage_spec)
        attachments_service.add(attachment)
      end
    end
```
if it does not look well, then i will change PR today as @gekorob proposed.

gekorob:
Hey @dsatsura no problem. I think that your solution can work for me. I made a PR to get all the contents from VM, so you can have the virtio_scsi.enabled support.
According to, just wait for that PR to be merged and then you can proceed.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
```diff
@@ -846,9 +850,10 @@ def calculate_max_memory(vm, memory)
     #
     def add_vm_disks(vm_service, disk_specs)
       storage_spec = disk_specs[:storage]
+      default_disk_spec = disk_specs[:default] || {}
```
gekorob: As far as I know add_vm_disks is called only by vm_reconfigure for now, but it's ok to put the empty hash.

gekorob:
There a couple of tests failing, that's because you have to put back VIRTIO in specs, 'cause it's the default in case of virtio_scsi not supported.

gekorob:
@dsatsura looks good to me, thanks a lot for your effort

PR #452, opened 2019-12-24
Native viewer support

gekorob:
Add native viewer support to allow the download of the connection file.
In case of multiple available consoles the "spice" one is consider the
preferred.
In case of headless vm (with no consoles available) an exception is
raised.

This is PR is a dependency for:
- https://github.com/ManageIQ/manageiq-ui-classic/pull/6574
- https://github.com/ManageIQ/manageiq/pull/19675

Related to BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1451594

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
@@ -196,6 +196,16 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices
       SystemConsole.launch_proxy_if_not_local(console_args, originating_server, host_address, host_port)
     end
 
+    def native_console_connection(vm)
+      vm.with_provider_object do |vm_service|
+        consoles = vm_service.graphics_consoles_service.list(:current => true)
+        return nil if consoles.empty?
+
+        console = select_graphics_console(consoles)
+        Base64.encode64(vm_service.graphics_consoles_service.console_service(console.id).remote_viewer_connection_file)
+      end
+    end
+
     def get_template_proxy(template, connection)
       TemplateProxyDecorator.new(
         connection.system_service.templates_service.template_service(template.uid_ems),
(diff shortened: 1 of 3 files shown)


gekorob:
@miq-bot add_label wip

gekorob:

gekorob:
@borod108 can I have your opinion please?
I add a "wip" label just because the way to utilize this functionality will depend on the choices in the ui. I'll open a PR also on the UI with my modifications.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
```diff
@@ -693,6 +703,12 @@ def collect_external_network_providers
 
     private
 
+    def select_graphics_console(consoles, pref_type = 'spice')
+      preferred = consoles.find { |c| c.protocol.downcase == pref_type }
+
```
borod108: Perhaps we should add a log message here in case a console of the preferred type was not found?

borod108: do we ever use pref_type?
and should we test it?

borod108:
@gekorob looks great! I had a really small comment, but it is not significant, let me know when you want me to merge this.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
gekorob: yeah you're right, I'll add some logs. Just wondering how to get logger reference in a simple way.

gekorob: No right know we're not using pref_type directly as a parameter. I afded it 'cause I supposed the preferred one is SPICE but leaving a way in the private method call to express another option. Not sure if to leave it as parameter or change it in a const...

gekorob:
Thanks @borod108 for the review, I'll try to fix the things you highlighted while waiting also feedback from UI on the use of the validate_native_viewer_support method

gekorob:
@miq-bot remove_label wip

gekorob:
@borod108 I changed method names from *native_viewer_console* to *native_console* as suggested in the UI review.
Introduced log in case *pref_type* console is not found and remove *pref_type* itself from the signature of *selected_graphics_console*, but used internally to choose in case of multiple consoles available.

Feel free to suggest modifications. 

I removed the wip label but, in case everything is ok for you, I think it's better to wait the review of UI modifications before merging.

borod108:
@gekorob great work! can you add a link to the UI PR in the description?
LGTM let me know when to merge.

gekorob:
@borod108 thanks a lot for reviewing and for the patience.
Sure, added references to the UI PR and the core one. For both this PR is a dependency.

gekorob:
@borod108, Whe you'll have time I think that we can merge this PR

PR #453, opened 2019-12-25
Allow user to set format and sparsity during provisioning and validate config

borod108:
We will now save the disk format as part of the refresh.
Will now allow the user to select disk_format
Will validate disk configuration.

Depends on: https://github.com/ManageIQ/manageiq-schema/pull/459
Depends on: https://github.com/ManageIQ/manageiq/pull/19849
Related to: https://github.com/ManageIQ/manageiq-ui-classic/pull/6692
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1726590

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
@@ -47,6 +47,9 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Cloning
   end
 
   def disk_format
+    format_from_dialog = get_option(:disk_format)
+    return format_from_dialog if format_from_dialog
+
     # If the datastore is not set, we will go with "block", because this will ensure
     # the result of the sparsity will be as set by the user.
     ds_type = 'block' unless dest_datastore
@@ -64,15 +67,15 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Cloning
   end
 
   def clone_type
-    get_option(:linked_clone).nil? ? clone_type_by_disk_format : clone_type_by_linked_clone
+    get_option(:linked_clone).nil? ? clone_type_by_disk_sparsity : clone_type_by_linked_clone
   end
 
   def clone_type_by_linked_clone
     get_option(:linked_clone) ? :linked : :full
   end
 
-  def clone_type_by_disk_format
-    get_option(:disk_format) == 'preallocated' ? :full : :linked
+  def clone_type_by_disk_sparsity
+    get_option(:disk_sparsity) == 'preallocated' ? :full : :linked
   end
 
   def template_clone_options
--- a/app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
@@ -88,7 +88,7 @@ module ManageIQ::Providers::Redhat::InfraManager::Provision::Cloning
   end
 
   def sparse_disk_value
-    case get_option(:disk_format)
+    case get_option(:disk_sparsity)
     when "preallocated" then false
     when "thin"         then true
     when "default"      then nil   # default choice implies inherit from template
--- a/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
@@ -219,13 +219,13 @@ class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisio
     src_template = load_ar_obj(src[:vm])
     if selected_linked_clone?
       s_id = src_template.storage_id
-      return storages.detect { |s| s.id == s_id }
+      return [storages.detect { |s| s.id == s_id }].compact
     end
 
     storages = storages.select do |storage|
       src_template.disks.all? do |disk|
         storage_type = storage_type_from_storage(storage)
-        validate_disk(disk, { :storage_type => storage_type })
+        validate_disk(disk, :storage_type => storage_type)
       end
     end
 
--- a/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
@@ -2,6 +2,8 @@ class ManageIQ::Providers::Redhat::InfraManager::ProvisionWorkflow < MiqProvisio
   include CloudInitTemplateMixin
   include SysprepTemplateMixin
 
+  include_concern "DialogFieldValidation"
+
   SYSPREP_TIMEZONES = {
     '001' => '(UTC-12:00) Dateline Standard Time',
     '002' => '(UTC-11:00) UTC-11',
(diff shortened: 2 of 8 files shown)


borod108:
@agrare @gekorob would love your feedback.

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -488,10 +488,11 @@ def hardware_disks(persister_hardware, disks)
           :location        => index.to_s,
           :size            => device.provisioned_size.to_i,
           :size_on_disk    => device.actual_size.to_i,
-          :disk_type       => device.sparse == true ? 'thin' : 'thick',
```
agrare: Let's continue to set `:disk_type` until we remove the column and either update the UI or provide a compatibility method for them to call.

borod108: done

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def storage_type
+    return nil unless dest_storage&.store_type
+
+    storage_type_from_storage(dest_storage)
+  end
+
+  DISK_CONFIGURATIONS = {
+    "file,raw,preallocated"  => true,
```
agrare: I think instead of constructing a comma separated string this would be slightly better with e.g.
```
storage_supported_format = {"file" => ["raw", "cow"], "block" => ["raw", "cow"]}
storage_supported_sparsity = {"file" => ["sparse"], "block" => ["preallocated", "sparse"]}

storage_supported_format[storage_type].include?(format) && storage_supported_sparsity[storage_type].include?(sparsity)
```

That way if this set grows we don't end up with an `N x M x P` set

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
```diff
@@ -47,6 +47,8 @@ def vm_clone_options
   end
 
   def disk_format
+    format_from_dialog =  get_option(:disk_format)
```
agrare: Really minor but there is an extra space after the `= `

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
```diff
@@ -72,7 +74,7 @@ def clone_type_by_linked_clone
   end
 
   def clone_type_by_disk_format
```
agrare: Should this method name change now that we are looking for sparsity not format?

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
```diff
@@ -212,14 +214,22 @@ def allowed_hosts_obj(_options = {})
 
   def allowed_storages(options = {})
     return [] if (src = resources_for_ui).blank?
-    result = super
+    storages = super
+    storages = storages.select { |s| s.storage_domain_type == "data" }
+    src_template = load_ar_obj(src[:vm])
+    if selected_linked_clone?
+      s_id = src_template.storage_id
+      return [storages.detect { |s| s.id == s_id }]
```
agrare: Can you `[].compact` this in case the storage detect doesn't find the storage for some reason and returns nil

agrare: Or continue to use `.select` which will return `[]` if nothing is found even though it is weird to use select to find by id

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
```diff
@@ -0,0 +1,86 @@
...
+    # is storage_type is not given we will assume it is "file" this way it will not block the validation,
+    # in reality the allowed_storages will actually run over each of the configurations with the available storages
+    # and will block the provisioning process if none of them is valid.
+    str_type = storage_type || "file"
+    DISK_CONFIGURATIONS["#{str_type},#{format},#{sparsity}"]
+  end
+
+  def dest_storage
+    ds_id = get_value(@values[:placement_ds_name])
+    return Storage.find(ds_id) if ds_id
```
agrare: Also minor, but `Storage.find(nil)` will raise the same `ActiveRecord::RecordNotFound` as if it were an invalid id.
That means the caller of this method doesn't have to deal with `dest_storage` raising an exception, returning nil, or returning a Storage record it just has to deal with two cases.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def storage_changed_for_disks?(vm)
+    return false unless dest_storage
```
agrare: Ah, if you want to return false if the storage isn't found then we should handle the `ActiveRecord::RecordNotFound` exception and return nil for any case

agrare:
Overall looks good just a few minor comments

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
borod108: I am not sure how to do that.
The three variables are not independent - for example the configuration "file,raw,preallocated" works, but "file,raw,sparse" does not, so cow works with file and preallocated works with file, but a combination of the two does not.

I do not think we can fit the information into two arrays with they keys being only one variable, and if we use complex keys seems to be even messier than just straight forward 3 dimensions.

But maybe I am just missing something...

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
borod108: done

app/models/manageiq/providers/redhat/infra_manager/provision/cloning.rb
borod108: oh tnx, missed that. done.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow.rb
borod108: right! done.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
borod108: cool, did not think about that.
It will now return nil or the storage.

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
borod108: done :)

app/models/manageiq/providers/redhat/infra_manager/provision_workflow/dialog_field_validation.rb
agrare: Oh okay hm, I'm good with this for now then

PR #454, opened 2020-01-03
Subclass InfraManager metrics capture

kbrock:
Ovirt is an infra metrics capture class. subclassing the proper parent

part of ManageIQ/manageiq#19684

NOTE: this one is not open/shut like the others.

The hierarchy is a little odd. Especially since there is a NetworkManager with network capture
Wondering if this is good enough, or if we need to get into modules and the like


Fryguy:
@agrare Please also review.

Issue #456, opened 2020-01-09
Can't refresh oVirt 4.2 provider


skateman:
When I try to refresh an oVirt 4.2 provider, it gives me the following error:
```
[----] E, [2020-01-09T14:48:37.324676 #61496:3719a0] ERROR -- : [NoMethodError]: undefined method `external_provider' for nil:NilClass  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2020-01-09T14:48:37.324856 #61496:3719a0] ERROR -- : /home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:44:in `id_of_external_network?'
/home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:519:in `block in vm_hardware_guest_devices'
/home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:511:in `each'
/home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:511:in `vm_hardware_guest_devices'
/home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:439:in `vm_hardware'
/home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-ovirt-c544028e4aea/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:400:in `block in vms'
[...]

PR #457, opened 2020-01-09
Fix for NICs that don't have a vnic_profile

agrare:
If a NIC doesn't have a vnic_profile looking up if it is an
external_network or not raises a NoMethodError trying to call
external_networks on a nil.

Fixes https://github.com/ManageIQ/manageiq-providers-ovirt/issues/456

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -41,7 +41,7 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
 
   def id_of_external_network?(network_id)
     network = collector.networks.detect { |net| net.id == network_id }
-    network.external_provider.present?
+    network&.external_provider.present?
   end
 
   def clusters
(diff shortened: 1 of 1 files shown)


agrare:
cc @skateman

app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -515,6 +515,8 @@ def vm_hardware_guest_devices(persister_hardware, vm, addresses, host)
       network = mac && networks.present? ? networks[mac] : nil
 
       vnic_profile_id = nic.dig(:vnic_profile, :id)
+      next if vnic_profile_id.nil?
```
agrare: @borod108 if this happens I wasn't sure if it was better to skip the NIC completely, or just not lookup the switch and lan below

borod108: I think it is not.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 8e466442f0966411666d79069458af1b0f62a493
Author: Boris Od <bodnopoz@redhat.com>
Date:   Thu Jan 9 20:21:06 2020 +0200

    Merge pull request #457 from agrare/fix_of_external_network_with_nil_id

    Fix for NICs that don't have a vnic_profile

    (cherry picked from commit 44a681f6c81d163308ae3da02368565963618309)

    https://bugzilla.redhat.com/show_bug.cgi?id=1767787
```

PR #458, opened 2020-01-10
Html5 console disabled by default

gekorob:
Html5 is not officially supported anymore in RHV.
We can disable it by default using a property in the settings file.
In this way the feature can be enabled at user own risk.

Related to BZ
https://bugzilla.redhat.com/show_bug.cgi?id=1668771

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm/remote_console.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm/remote_console.rb
@@ -1,12 +1,17 @@
 class ManageIQ::Providers::Redhat::InfraManager::Vm
   module RemoteConsole
     def console_supported?(type)
-      %w(SPICE VNC NATIVE).include?(type.upcase)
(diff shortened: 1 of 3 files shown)


gekorob:

gekorob:
@agrare @borod108 I tried to develop a solution to remove the html5 console but letting the user the possibility to re-enable it changing the value into settings.
I don't if there are rules to define parameters in settings or if it's an acceptable solution.
What do you think?

borod108:
@gekorob LGTM, will merge once tests are green (I know it has nothing to do with this PR, but still just to follow protocol...)

gekorob:
@borod108, yes no problem. It's correct.

simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-providers-ovirt/pull/514

simaishi:
The ivanchuk backport was reverted.

PR #459, opened 2020-01-10
Avoid a new OvirtServices instantiation

gekorob:
Simple refactor to avoid a new instantiation using the factory method

@miq-bot add_label refactoring


borod108:
LGTM

PR #462, opened 2020-01-30
Don't check for Datacenter class

agrare:
The hash isn't an ActiveRecord object yet so we can't use
f.kind_of?(Datacenter)

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
@@ -78,8 +78,7 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParser
   end
 
   def self.parse_new_target(full_data, message, ems, event_type)
-    folders = parse_new_folders(full_data.data_center)
-    dc      = folders.detect { |f| f.kind_of?(Datacenter) }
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
```diff
@@ -78,8 +78,7 @@ def self.ems_ref_from_object_in_event(data)
   end
 
   def self.parse_new_target(full_data, message, ems, event_type)
-    folders = parse_new_folders(full_data.data_center)
-    dc      = folders.detect { |f| f.kind_of?(Datacenter) }
+    dc, vm_folder, host_folder = parse_new_folders(full_data.data_center)
```
Fryguy: optionally you can also do...
```ruby
dc, *folders = parse_new_folders(full_data.data_center)
...
:folders => [dc, *folders]
```

agrare: :+1: done

agrare:
@borod108 @mwperina master has been red since Jan 21st

chessbyte:
Merging with test failures as they are not related to this PR.

PR #466, opened 2020-02-05
Ems metrics collector section should remain

gekorob:
Without ems_metrics_collector_worker_redhat section the server is unable to
start

See PR:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/464


gekorob:
@borod108 test were passing when I tested the PR referred here, but experimenting with the real environment I realized that server was not able to start with the following runtime error:
*Missing config section ems_metrics_collector_worker_redhat (RuntimeError)*
Could you please review?

PR #467, opened 2020-02-06
OvirtSDK4 objects conversion to Hash

gekorob:
'OvirtSDK4::Event's were stored in the full_data as object since the
beginnig. This was causing problems in the dequeue process

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
@@ -46,7 +46,7 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParser
         :message             => event.description,
         :timestamp           => event.time,
         :username            => username,
-        :full_data           => event,
+        :full_data           => ovirtobj_to_hash(event),
         :ems_id              => ems_id,
         :vm_ems_ref          => vm_ref,
         :host_ems_ref        => ems_ref_from_object_in_event(event.host),
@@ -77,7 +77,8 @@ module ManageIQ::Providers::Redhat::InfraManager::EventParser
     ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(data.href)
   end
 
-  def self.parse_new_target(full_data, message, ems, event_type)
+  def self.parse_new_target(full_data_hash, message, ems, event_type)
+    full_data = hash_to_ostruct(full_data_hash)
     dc, *folders = parse_new_folders(full_data.data_center)
 
     cluster = parse_new_cluster(ems, full_data.cluster, dc)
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
```diff
@@ -1,3 +1,5 @@
+require 'ostruct'
```
agrare: Can you move this require down to where you're using OpenStruct? https://github.com/ManageIQ/manageiq-providers-ovirt/pull/467/files#diff-52fa558b219d05b5565f5fc8ea42e88cR214

We try to keep global requires down so code that doesn't need this doesn't have to load it for memory usage reasons

app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
```diff
@@ -206,4 +209,27 @@ def self.parse_new_resource_pool(cluster)
...
+      end
+    )
+  end
+
+  def self.ovirtobj_to_hash(obj)
+    obj.instance_variables.each_with_object({}) do |k, h|
+      val = obj.instance_variable_get(k)
+
+      h[k.to_s[1..-1]] = case val
+                         when OvirtSDK4::Identified
```
agrare: I assume this is the base class for anything else?

app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
gekorob: Yes sure

app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
```diff
@@ -206,4 +209,27 @@ def self.parse_new_resource_pool(cluster)
...
+        memo[key] = val.kind_of?(Hash) ? hash_to_ostruct(val) : val
+      end
+    )
+  end
+
+  def self.ovirtobj_to_hash(obj)
+    obj.instance_variables.each_with_object({}) do |k, h|
+      val = obj.instance_variable_get(k)
+
+      h[k.to_s[1..-1]] = case val
```
agrare: Just curious what does `k.to_s` look like that you had to strip out the first char?

gekorob: for the '@'

agrare: Ah of course

app/models/manageiq/providers/redhat/infra_manager/event_parser.rb
gekorob: Yes according to the docs is the base for every common OvirtSDK4 obj (I assume is the base for all that we can find in events)
https://www.rubydoc.info/gems/ovirt-engine-sdk/4.3.0/OvirtSDK4/Identified

agrare: :+1:

PR #471, opened 2020-02-18
Add location for localfs storage

gekorob:
ems_ref is explicity set as location attribute for LOCALFS
storagedomains.
In the upstream master this was also the implicit value that the inventory_refresh is setting in case of location nil, but for hammer and ivanchuck this is necessary to fix the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1798795

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -71,7 +71,11 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
   def storagedomains
     collector.storagedomains.each do |storagedomain|
       storage_type = storagedomain.dig(:storage, :type).upcase
-      location = if storage_type == 'NFS' || storage_type == 'GLUSTERFS'
+      ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(storagedomain.try(:href))
+      location = case storage_type
+                 when 'LOCALFS'
+                   ems_ref
+                 when 'NFS', 'GLUSTERFS'
                    "#{storagedomain.dig(:storage, :address)}:#{storagedomain.dig(:storage, :path)}"
                  else
                    storagedomain.dig(:storage, :volume_group, :id)
@@ -81,8 +85,6 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
       used        = storagedomain.try(:used).to_i
       total       = free + used
       committed   = storagedomain.try(:committed).to_i
-
-      ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(storagedomain.try(:href))
 
       persister.storages.find_or_build(ems_ref).assign_attributes(
         :ems_ref             => ems_ref,


app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -71,7 +71,11 @@ def clusters
...
       storage_type = storagedomain.dig(:storage, :type).upcase
-      location = if storage_type == 'NFS' || storage_type == 'GLUSTERFS'
+      ems_ref = ManageIQ::Providers::Redhat::InfraManager.make_ems_ref(storagedomain.try(:href))
+      location = case storage_type
+                 when 'LOCALFS'
+                   ems_ref
+                 when 'NFS', 'GLUSTERFS'
                    "#{storagedomain.dig(:storage, :address)}:#{storagedomain.dig(:storage, :path)}"
                  else
                    storagedomain.dig(:storage, :volume_group, :id)
```
agrare: @borod108 do you know if this `else` case is really for all the other types of storage, or was it iSCSI and/or local block storage?

I'm wondering if we make `ems_ref` the "else" because it is guaranteed to at least work

borod108: @agrare the "else" really does seem to work work for FCP as well (tested manually)
but in case we do not care about the location as much as before (which seems to be true with the latest changes) I would just change all of it to the ems_ref....

agrare: Yeah we should try to keep the location the same for existing storages though.  I'm good with this then.

agrare: If we can follow up with a more complete list of storage types and the ones where that volume_group.id doesn't work we can use href.

gekorob: According to the ovirt-engine api model we have the storage types described here:
https://ovirt.github.io/ovirt-engine-api-model/4.3/#types/storage_type

If I can say my opinion, to both match the intent of keeping the same location for existing storages and be "safe" in case of not well explored storages we can do something like the following:

 ```
location = case storage_type
                 when 'NFS', 'GLUSTERFS'
                   "#{storagedomain.dig(:storage, :address)}:#{storagedomain.dig(:storage, :path)}"
                 when 'ISCSI', 'FCP'
                   storagedomain.dig(:storage, :volume_group, :id)
                 else
                   ems_ref
```
Because I tried to query a glance storage type and it seems to me that it does not have more specific attributes than the LOCALFS one.

Comment (agrare) [reply app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb]: :+1: that looks great, lets backport this simple fix and we can do this in a followup

Comment (gekorob) [reply app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb]: Ok perfect thx

**Comment by simaishi** [comment]
Ivanchuk backport details:

```
$ git log -1
commit dd4aeab2b2b898f9b0494910ea7d47fef9688861
Date:   Wed Feb 19 08:34:23 2020 -0500

    Merge pull request #471 from gekorob/localfs_location_fix

    Add location for localfs storage

    (cherry picked from commit aceb2019b89a8bdffa7f45bf5e52b9a6528cf54f)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805909
```

PR #474, opened 2020-02-19
[HAMMER] We need all VM details

gekorob:
While reconfiguring disk attachments for a VM,
we need the virtio_scsi support flag to set the disk interface
accordingly, in case it's not specified.

Backport of #451

(cherry picked from commit ad5336c)

Related to BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1803775


gekorob:
No new releases expected for 5.10 in the near future. We'll reopen this one in case needed

PR #475, opened 2020-02-19
[HAMMER] Enable to set virtio_scsi disk attachment interface

gekorob:
Fix issue when disk_add :interface option from VmReconfigure request was not passed to DiskAttachmentBuilder

virtio_scsi is default interface option in oVirt/RHV UI, so make it default in ManageIQ as well.

(cherry picked from commit d8d3bce)

Backport of #445

Related to BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1803775

Depends on:
#474


gekorob:
No new releases expected for 5.10 in the near future. We'll reopen this one in case needed

PR #476, opened 2020-02-20
Require manageiq-smartstate in VmOrTemplate::Scanning

agrare:
We need to require the manageiq-smartstate gem in the scanning class to
prevent MiqServer#scan_metadata from the following error:

`uninitialized constant ManageIQ::Smartstate`


borod108:
LGTM

agrare:
Thanks guys!

PR #477, opened 2020-02-25
Add network reconfiguration

gekorob:
Vm reconfiguration extended to NICs, allowing
user to add, remove and edit VM network interfaces

Depends on: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/480

Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1751151

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
@@ -29,7 +29,8 @@ module ManageIQ::Providers::Redhat::InfraManager::Vm::Reconfigure
       "memoryMB"          => (task_options[:vm_memory].to_i if task_options[:vm_memory]),
       "numCPUs"           => (task_options[:number_of_cpus].to_i if task_options[:number_of_cpus]),
       "disksRemove"       => task_options[:disk_remove],
-      "disksAdd"          => (spec_for_added_disks(task_options[:disk_add]) if task_options[:disk_add])
+      "disksAdd"          => (spec_for_added_disks(task_options[:disk_add]) if task_options[:disk_add]),
+      "networkAdapters"   => spec_for_network_adapters(task_options)
     }
   end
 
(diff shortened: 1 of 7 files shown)


app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
```diff
@@ -39,4 +40,103 @@ def spec_for_added_disks(disks)
...
+    {
+      :edit   => (network_adapters_edit(options[:network_adapter_edit]) if options[:network_adapter_edit]),
+      :add    => (network_adapters_add(options[:network_adapter_add]) if options[:network_adapter_add]),
+      :remove => (network_adapters_remove(options[:network_adapter_remove]) if options[:network_adapter_remove])
+    }.compact
+  end
+
+  def network_adapters_add(adapters)
+    nic_names = hardware.nics.pluck(:device_name)
+    switch_ids = HostSwitch.where(:host_id => host.id).pluck(:switch_id)
```
agrare: This can be just `.where(:host => host)` no need to have it be `host.id`

app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
```diff
@@ -39,4 +40,103 @@ def spec_for_added_disks(disks)
...
+  def find_lan_by_name(network_name, switch_ids)
+    lan_name, ext_switch_name = network_name.split('/').collect(&:strip)
+
+    switch_ids = ext_switch_ids_by_name(ext_switch_name) if ext_switch_name
+
+    Lan.find_by(:name => lan_name, :switch_id => switch_ids)
+  end
+
+  def ext_switch_ids_by_name(switch_name)
+    ext_management_system.external_distributed_virtual_switches.where(:name => switch_name).pluck(:id)
```
agrare: @gekorob is the name of the DVS unique across the whole EMS?

agrare: Aka is it possible to have two Lans with the same name on two switches with the same name

gekorob: @agrare in RHV is not possible to add a new logical network with the same name of another logical network in the same datacenter

agrare: :+1:

app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
gekorob: Yeah right, sorry. I'll fix immediately

agrare:
Looks good overall, I'll let the RHV team review the actual ovirt sdk side though

app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
gekorob: @agrare I made some experiments and your first doubt was correct. I mean what I wrote about RHV is correct, but within the same EMS we can have more datacenters.
The problem is that we have external_distributed_virtual_switches connected to the ems and that, I think, is not correctly modelling the RHV situation

gekorob: @borod108 is there a way to get external_distributed_virtual_lans for the datacenter of the vm under reconfiguration (without asking to the provider)? or am I missing something?

agrare: Okay so hosts under a datacenter are not connected to switches from another datacenter?

That was why I suggested just creating host_switches join table records for every host to the distributed switches so that the rest of the modeling would "Just work" but I think there were some issues with that that maybe @borod108 remembers.  Something to do with targeted refresh but I think this could be overcome.

gekorob: Will it be possible to have a join table between datacenter and switches? In particular to have external_distributed_virtual_switches connected to datacenter?

agrare: Okay so now that we have a relationship between external switches and their owning datacenter we can use that here right?

gekorob: Yes I'm about to push the commits with tests

gekorob: I was testing also on a RHV env

agrare: :heart: awesome

app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
```diff
@@ -39,4 +40,103 @@ def spec_for_added_disks(disks)
...
+  def find_lan_by_name(network_name, switch_ids)
+    lan_name, ext_switch_name = network_name.split('/').collect(&:strip)
+
+    switch_ids = ext_switch_ids_by_name(ext_switch_name) if ext_switch_name
+
+    Lan.find_by(:name => lan_name, :switch_id => switch_ids)
+  end
+
+  def ext_switch_ids_by_name(switch_name)
+    parent_datacenter.external_distributed_virtual_switches.select { |s| s.name == switch_name }.map(&:id)
```
agrare: :+1: nice I like how this came out

gekorob: Thanks to your helpful tips!

simaishi:
Adding `jansa/yes` as https://github.com/ManageIQ/manageiq-providers-ovirt/pull/478 is `jansa/yes` and depends on this PR.

simaishi:
Jansa backport details:

```
$ git log -1
commit 6191474d97a7788634bd9e6ed87e0db3d841a7cd
Date:   Thu Mar 5 10:02:41 2020 -0500

    Merge pull request #477 from gekorob/rfe/vm_network_reconf

    Add network reconfiguration

    (cherry picked from commit 7b64737ed9ea1e366fcc3974075995a56478b031)

    https://bugzilla.redhat.com/show_bug.cgi?id=1751151
```

PR #478, opened 2020-02-25
Available vlans for network reconfigure

gekorob:
Returns a list of available Vm vlans including the external ones

Depends on: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/477
Depends on: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/480

Related to: https://github.com/ManageIQ/manageiq-ui-classic/pull/6710
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1751151

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
@@ -24,24 +24,13 @@ module ManageIQ::Providers::Redhat::InfraManager::Vm::Reconfigure
   end
 
   def available_vlans
-    vlans = []
-
-    switch_ids = HostSwitch.where(:host => host).pluck(:switch_id)
-    Lan.where(:switch_id => switch_ids).each do |lan|
-      vlans << lan.name
-    end
+    vlans = host.lans.pluck(:name)
 
     vlans.sort.concat(available_external_vlans)
   end
 
   def available_external_vlans
-    ext_vlans = []
-    ems = ext_management_system
-
-    ext_switch_ids = ems.external_distributed_virtual_switches.pluck(:id)
-    ems.external_distributed_virtual_lans.where(:switch_id => ext_switch_ids).each do |ext_lan|
-      ext_vlans << "#{ext_lan.name}/#{ext_lan.switch.name}"
-    end
+    ext_vlans = ext_management_system.external_distributed_virtual_lans.map { |lan| "#{lan.name}/#{lan.switch.name}" }
 
     ext_vlans.sort
   end
--- a/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
@@ -30,7 +30,7 @@ module ManageIQ::Providers::Redhat::InfraManager::Vm::Reconfigure
   end
 
   def available_external_vlans
-    ext_vlans = ext_management_system.external_distributed_virtual_lans.map { |lan| "#{lan.name}/#{lan.switch.name}" }
+    ext_vlans = parent_datacenter.external_distributed_virtual_lans.map { |lan| "#{lan.name}/#{lan.switch.name}" }
 
     ext_vlans.sort
   end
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
```diff
@@ -23,6 +23,29 @@ def max_memory_mb
     2.terabyte / 1.megabyte
   end
 
+  def available_vlans
+    vlans = []
+
+    switch_ids = HostSwitch.where(:host_id => host_id).pluck(:switch_id)
+    Lan.where(:switch_id => switch_ids).each do |lan|
+      vlans << lan.name
+    end
```
agrare: Can this just be `host.lans.pluck(:name)` ?

app/models/manageiq/providers/redhat/infra_manager/vm/reconfigure.rb
```diff
@@ -23,6 +23,29 @@ def max_memory_mb
...
+  end
+
+  def available_external_vlans
+    ext_vlans = []
+    ems = ext_management_system
+
+    ext_switch_ids = ems.external_distributed_virtual_switches.pluck(:id)
+    ems.external_distributed_virtual_lans.where(:switch_id => ext_switch_ids).each do |ext_lan|
+      ext_vlans << "#{ext_lan.name}/#{ext_lan.switch.name}"
+    end
```
agrare: Same question here, `external_distributed_virtual_lans` is already `:through => :external_distributed_virtual_switches` so I think you're re-implementing the rails association here without the caching :)

Try just `ems.external_distributed_virtual_lans.map { |lan| "#{lan.name}/#{lan.switch.name}" }`

gekorob: I used the same style that was used in the manageiq-ui-classic. I was not sure if it was better to use rails relations involving more objects or use the pluck.

agrare: @gekorob can you link to the code that you found in UI classic?  Curious on the history there

agrare: @gekorob can you link to the code that you found in UI classic?  Curious on the history there

gekorob: @agrare yes
https://github.com/ManageIQ/manageiq-ui-classic/blob/e7e66e5014f25fc2c2e0ec9a1a115602dea93c57/app/controllers/mixins/actions/vm_actions/reconfigure.rb#L214-L226
I think the purpose was also to use filter access on Roles

gekorob: I tried to propose a simple way to delegate the list of vlans to the provider if it implements the method "available_vlans"
https://github.com/gekorob/manageiq-ui-classic/blob/2122a372aaa87ddca3abe5ea43b51cba53aa9518/app/controllers/mixins/actions/vm_actions/reconfigure.rb#L22

Maybe not the best thing, but similar what was done for adapters_name and currently implemented in the vmware provider I think
https://github.com/gekorob/manageiq-ui-classic/blob/2122a372aaa87ddca3abe5ea43b51cba53aa9518/app/controllers/mixins/actions/vm_actions/reconfigure.rb#L24

sigbjornaib:
Will there be possible to filter the network reconfiguration in a Role? Today there is only an option to allow or disallow "Reconfigure VMs". 

Giving access to end users to change the amount of CPU, RAM and add Disk, is not same the as allowing them to change VLAN which implies allowing them to change the Firewall Zone of the VM.

gekorob:
@sigbjornaib I'm not deep into Role authorization, but I think that with some roles the user is not allowed to reconfigure Vms. I don't know if it is possible to choose which reconfiguration option to allow.
At plugin level AFAIK we can only decide to support (and implement) or not a particular reconfiguration capability

simaishi:
Jansa backport details:

```
$ git log -1
commit d1c27e3db1e3803dae8cf81dde6ff9b364ec9a16
Date:   Tue Mar 10 09:27:47 2020 -0400

    Merge pull request #478 from gekorob/available_vlans

    Available vlans for network reconfigure

    (cherry picked from commit ce43a966e189502c101c5b5d815112898cf69c5c)

    https://bugzilla.redhat.com/show_bug.cgi?id=1751151
```

PR #479, opened 2020-03-03
Fix constant redefinition warnings in specs

djberg96:
Currently if you run the specs locally you will see the following warnings:

```
spec/models/manageiq/providers/redhat/infra_manager/refresher/refresher_graph_target_vm_spec.rb:16: warning: already initialized constant COUNTED_MODELS
spec/models/manageiq/providers/redhat/infra_manager/refresher/refresher_graph_target_template_spec.rb:11: warning: previous definition of COUNTED_MODELS was here
spec/models/manageiq/providers/redhat/infra_manager/refresher/refresher_huge_async_spec.rb:15: warning: already initialized constant ORIG_YML_PATH
spec/models/manageiq/providers/redhat/infra_manager/refresher/refresh_recording_modifier_spec.rb:12: warning: previous definition of ORIG_YML_PATH was here
```
~~This PR stubs out the constants properly, with the bonus that these warnings go away.~~

Edit: We can just use `let` here.

### DIFF
--- a/spec/models/manageiq/providers/redhat/infra_manager/refresher/refresh_recording_modifier_spec.rb
+++ b/spec/models/manageiq/providers/redhat/infra_manager/refresher/refresh_recording_modifier_spec.rb
@@ -4,16 +4,17 @@ require_relative 'ovirt_refresher_spec_common'
 describe ManageIQ::Providers::Redhat::InfraManager::Refresher do
   include OvirtRefresherSpecCommon
 
+  let(:orig_yml_path) { 'spec/vcr_cassettes/manageiq/providers/redhat/infra_manager/refresh/ovirt_sdk_refresh_recording_for_mod.yml'.freeze }
+
   before(:each) do
     init_defaults
     init_connection_vcr
-    stub_const('ORIG_YML_PATH', 'spec/vcr_cassettes/manageiq/providers/redhat/infra_manager/refresh/ovirt_sdk_refresh_recording_for_mod.yml'.freeze)
   end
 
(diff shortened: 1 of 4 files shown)


djberg96:
@miq-bot add_label test, refactoring

agrare:
Hmm I feel like there's a better way to fix this

spec/models/manageiq/providers/redhat/infra_manager/refresher/refresh_recording_modifier_spec.rb
```diff
@@ -7,10 +7,9 @@
   before(:each) do
     init_defaults
     init_connection_vcr
+    stub_const('ORIG_YML_PATH', 'spec/vcr_cassettes/manageiq/providers/redhat/infra_manager/refresh/ovirt_sdk_refresh_recording_for_mod.yml'.freeze)
   end
 
-  ORIG_YML_PATH = 'spec/vcr_cassettes/manageiq/providers/redhat/infra_manager/refresh/ovirt_sdk_refresh_recording_for_mod.yml'.freeze
```
agrare: Does this even need to be a constant?

djberg96: Maybe you're right. I thought it was something hooked into, but now that I look closer I think it's just being used as a global.

djberg96:
@agrare Ok, updated to use `let` instead of constants. I did have to change a couple variable names to prevent overwriting, but otherwise it was pretty straightforward.

PR #482, opened 2020-03-05
Provision workflow fix

gekorob:
The external vlans must be retrieved from the cluster's parent
datacenter

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
@@ -1018,7 +1018,7 @@ module ManageIQ::Providers::Redhat::InfraManager::OvirtServices
 
     def private_load_allowed_networks(vlans, uid_ems_cluster)
       profiles = local_get_vnic_profiles_in_cluster(uid_ems_cluster)
-      profiles.merge!(local_external_vnic_profiles)
+      profiles.merge!(local_external_vnic_profiles(uid_ems_cluster))
       profiles.each do |profile, profile_network|
         vlans[profile.uid_ems] = "#{profile.name} (#{profile_network.name})"
       end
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
```diff
@@ -1054,10 +1055,14 @@ def local_external_vnic_profiles
     end
 
     def all_networks_as_switches_in_cluster(uid_ems_cluster)
-      cluster = EmsCluster.find_by(:uid_ems => uid_ems_cluster)
+      cluster = cluster_by_uid_ems(uid_ems_cluster)
       cluster&.switches
     end
 
+    def cluster_by_uid_ems(uid_ems)
+      EmsCluster.find_by(:uid_ems => uid_ems)
```
agrare: Do you have access to the ems_id from here?  We shouldn't be looking up records by just uid_ems since they aren't guaranteed to be unique.

Ideally this would be `ext_management_system.ems_clusters.find_by(:uid_ems => uid_ems_cluster)`

agrare: I realize this is just a move but it is a good chance to fix an existing bug

gekorob: I agree, I refactored to avoid duplication. I'll change it right now.

app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
```diff
@@ -1040,9 +1040,10 @@ def local_get_vnic_profiles_in_cluster(uid_ems_cluster)
       cluster_profiles
     end
 
-    def local_external_vnic_profiles
+    def local_external_vnic_profiles(uid_ems_cluster)
       external_profiles = {}
-      profiles = ext_management_system.external_distributed_virtual_lans
+      cluster = cluster_by_uid_ems(uid_ems_cluster)
+      profiles = cluster.parent_datacenter.external_distributed_virtual_lans
       external_networks = ext_management_system.external_distributed_virtual_switches
```
gekorob: @agrare, I think that also here we should get `external_distributed_virtual_switches` from the datacenter, right?

app/models/manageiq/providers/redhat/infra_manager/ovirt_services.rb
agrare: :+1: looks good now

PR #487, opened 2020-03-30
[WIP] add event target parser start

d-m-u:
what it says on the tin

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/event_target_parser.rb
@@ -1,0 +1,50 @@
+class ManageIQ::Providers::Redhat::InfraManager::EventTargetParser
+  attr_reader :ems_event
+
+  # @param ems_event [EmsEvent] EmsEvent object
+  def initialize(ems_event)
+    @ems_event = ems_event
+  end
+
+  # Parses all targets that are present in the EmsEvent given in the initializer
+  #
+  # @return [Array] Array of InventoryRefresh::Target objects
+  def parse
+    parse_ems_event_targets(ems_event)
+  end
+
+  private
+
+  # Parses list of InventoryRefresh::Target out of the given EmsEvent
+  #
+  # @param event [EmsEvent] EmsEvent object
+  # @return [Array] Array of InventoryRefresh::Target objects
+  def parse_ems_event_targets(event)
+    target_collection = InventoryRefresh::TargetCollection.new(:manager => event.ext_management_system, :event => event)
+
+    if ems_event.event_type.start_with?("orchestration.stack")
+      collect_orchestration_stack_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("image.")
+      collect_image_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("aggregate.")
+      collect_host_aggregate_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("keypair")
+      collect_key_pair_references!(target_collection, ems_event)
+    end
+
+    target_collection.targets
+  end
+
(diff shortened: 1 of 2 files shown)


gekorob:
@d-m-u, @agrare is this related to https://github.com/ManageIQ/manageiq/issues/19999 ?

d-m-u:
@gekorob yeah, I just wanted it to get reviewed before I linked it.

app/models/manageiq/providers/redhat/infra_manager/event_target_parser.rb
```diff
@@ -0,0 +1,90 @@
...
+  #
+  # @param event [EmsEvent] EmsEvent object
+  # @return [Array] Array of InventoryRefresh::Target objects
+  def parse_ems_event_targets(event)
+    target_collection = InventoryRefresh::TargetCollection.new(:manager => event.ext_management_system, :event => event)
+
+    # there's almost always a tenant id regardless of event type
+    collect_identity_tenant_references!(target_collection, ems_event)
+
+    if ems_event.event_type.start_with?("compute.instance")
```
agrare, line 28: Does RHV has events that start with `"compute.instance"` ?  I don't see any of these events in https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class

It looks like they were copied from OpenStack

d-m-u: oh they were

agrare: Okay lets remove this set of checks as well as https://github.com/ManageIQ/manageiq-providers-ovirt/pull/487/files#diff-26c9135e0247d314dab264c48e4a728bR26 and all these https://github.com/ManageIQ/manageiq-providers-ovirt/pull/487/files#diff-26c9135e0247d314dab264c48e4a728bR51-R89

app/models/manageiq/providers/redhat/infra_manager/event_target_parser.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    if ems_event.event_type.start_with?("orchestration.stack")
+      collect_orchestration_stack_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("image.")
+      collect_image_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("aggregate.")
+      collect_host_aggregate_references!(target_collection, ems_event)
+    elsif ems_event.event_type.start_with?("keypair")
+      collect_key_pair_references!(target_collection, ems_event)
+    end
```
agrare, line 33: RHV doesn't have any of these

spec/models/manageiq/providers/redhat/infra_manager/event_target_parser_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
+    @ems                 = FactoryBot.create(:ems_redhat, :zone => zone)
+
+    allow_any_instance_of(EmsEvent).to receive(:handle_event)
+    allow(EmsEvent).to receive(:create_completed_event)
+  end
+
+  context "Redhat Event Parsing" do
+    it "parses identity.project events" do
+      ems_event = create_ems_event("identity.project.create.end", "project_id" => "tenant_id_test")
```
agrare, line 12: This is an openstack event isn't it?

agrare:
@d-m-u I think what we should do is start with actual RHV events in the specs and then get the EventTargetParser handling these.  Right now you have openstack events and an openstack parser so the specs are green even though this wouldn't actually work.

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-ovirt/pull/515

PR #501, opened 2020-06-16
Add cpu type and speed to vm hardware info

djberg96:
Attempting to add cpu type and speed to the vm hardware for the Ovirt provider.

Draft for now since the information appears to always be empty.

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -444,8 +444,10 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
       :annotation           => vm.try(:description),
       :cpu_cores_per_socket => cpu_cores,
       :cpu_sockets          => cpu_socks,
+      :cpu_speed            => vm.cpu.speed,
       :cpu_total_cores      => cpu_cores * cpu_socks,
-      :memory_mb            => vm.memory / 1.megabyte
+      :cpu_type             => vm.cpu.type,
+      :memory_mb            => vm.memory / 1.megabyte,
     )
 
     hardware_disks(persister_hardware, disks)


djberg96:
Using the SDK directly, I'm seeing stuff like this as a response for a vm.cpu:

```
#<OvirtSDK4::Cpu:0x00007fe8d8025b70
  @href=nil,
  @architecture="x86_64",
  @cores=nil,
  @cpu_tune=nil,
  @level=nil,
  @mode=nil,
  @name=nil,
  @speed=nil,
  @topology=#<OvirtSDK4::CpuTopology:0x00007fe8d80257d8 @href=nil, @cores=2, @sockets=2, @threads=1>, @type=nil>
```
Why are most fields blank? Is this is an issue with the SDK? Or does the raw response not include this information? Still digging.

gekorob:
Hi @djberg96, as far as I know the speed and type parameters for the cpu object are populated for the host
```
host = conn.follow_link(vm.host)
host.cpu

#<OvirtSDK4::Cpu:0x0000000001b5ad00 
@href=nil, 
@architecture=nil, 
@cores=nil, @cpu_tune=nil, @level=nil, @mode=nil, 
@name="Intel Xeon Processor (Skylake)", 
@speed=2100.0, 
@topology=#<OvirtSDK4::CpuTopology:0x0000000001b59798 @href=nil, 
@cores=4, @sockets=1, @threads=1>, 
@type="Intel Skylake Server IBRS SSBD Family">
```

djberg96:
@gekorob Makes sense, if it's a guest VM on a host, it won't have its own cpu information.

Alright, I may tinker with overloading the vm.rb file accessors instead.

gekorob:
@djberg96 Thanks for the effort, in case you need anything, just ask

djberg96:
BZ was closed, so closing this. The CPU information displayed in the RHV UI comes from the host or the cluster, though even then it's just the family.

PR #504, opened 2020-06-22
Collect tools status information

djberg96:
~~This is a work in progress for https://bugzilla.redhat.com/show_bug.cgi?id=1846624~~

As per discussions on gitter, I was originally just going to look at a the applications list for a given VM and check for the "ovirt-guest-agent". However, I noticed that we collect more information for SCVMM:

https://github.com/ManageIQ/manageiq-providers-scvmm/blob/master/app/models/manageiq/providers/microsoft/infra_manager/parser_mixin.rb#L107-L116

That lead me to look at the schema to see if we had a table we could link to instead of storing information directly on the vms table, and I found the `guest_applications` table which has a relationship to either the `vms` table or the `hosts` table.

So, for now the plan is to get information into the `guest_applications` table, then use `vm.guest_applications.any?{ |ga| ga.name =~ /guest-agent/ }` to set `tools_status`.

Draft in progress for now. I'm not sure how to handle `guest_applications` in the Persister since that table doesn't have an `ems_ref` field.

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/collector.rb
+++ b/app/models/manageiq/providers/redhat/inventory/collector.rb
@@ -157,7 +157,7 @@ class ManageIQ::Providers::Redhat::Inventory::Collector < ManageIQ::Providers::I
     end
   end
 
-  def collect_guest_applications(vm)
+  def collect_vm_guest_applications(vm)
     manager.with_provider_connection do |connection|
       connection.follow_link(vm.applications)
     end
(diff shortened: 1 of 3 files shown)


gtanzillo:
This was merged in PR #505

djberg96:
@gtanzillo Well, this was a more elaborate approach I was working on for the future, where we actually store the application list in the `guest_applications` table. But, it's not required, no.

Issue #508, opened 2020-07-03
Missing asynchronous context-free credential validation for DDF


skateman:
There's no `self.verify_credentials` for the add provider form, also the already existing validation operates with DB models, so it's basically the same problem as in https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/387 but here for the default and the metrics endpoint.

@miq-bot assign @agrare 
Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

agrare:
Hey @skateman actually there is one it is just in [`InfraManager::ApiIntegrations`](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb#L209-L233).  I had to update it to match what you have in https://github.com/ManageIQ/manageiq-providers-ovirt/pull/509 so I'll push the changes to your branch but verification and adding the provider works, going to close this

PR #509, opened 2020-07-03
Adjusted DDF schema and validation for creating new providers

skateman:
**Before:**
Screenshot from 2020-07-03 11-23-36

**After:**
Screenshot from 2020-07-03 11-23-11

Depends on: https://github.com/ManageIQ/manageiq-providers-ovirt/issues/508
Parent issue: https://github.com/ManageIQ/manageiq/issues/18818


skateman:
@agrare good to go

PR #510, opened 2020-07-15
set transform_vm system flag to true

d-m-u:
This is the only system dialog we have left and should have something to differentiate it. 
yes, i know it's kind of moot since we only have one, but it's functionality that made sense and i'd like to close the issue assigned to me (https://github.com/ManageIQ/manageiq/issues/15144).

see the related https://github.com/ManageIQ/manageiq-schema/pull/471

### DIFF
--- a/content/service_dialogs/transform-vm.yml
+++ b/content/service_dialogs/transform-vm.yml
@@ -26,6 +26,7 @@
         display_method:
         display_method_options: {}
         required: false
+        system: true
         required_method:
         required_method_options: {}
         default_value: 'migration_group'


content/service_dialogs/transform-vm.yml
```diff
@@ -26,6 +26,7 @@
         display_method:
         display_method_options: {}
         required: false
+        system: true
```
bdunne, line 29: I think this should be forced in core `Dialog.seed` so that people don't need to remember to do it here. https://github.com/ManageIQ/manageiq-schema/pull/471#issuecomment-658939959

d-m-u: I disagree. I think this _should_ be here. @tinaafitz would you like to comment, please?

bdunne: @Fryguy Can you comment on how the `system` flag should be set during seed?

Fryguy: I guess the question is...will every single entry in any shipped plugin need to be system:true?  If so, I agree with @bdunne this should be more of a low-level seeding thing.  Though I am ok with this as a one-off fix

d-m-u: > will every single entry in any shipped plugin need to be system:true

If you see the schema PR that Brandon linked to, this is the only one of these we have left, and Tina did mention something about it being deprecated.

tinaafitz: I don't know the road map, so I can only speak to the current functionality.

1. Setting the yaml file`system` flag:
 a. Gives providers the ability to provide OOTB non-system Service dialogs.
 b. Gives users the ability to copy their own Service dialogs to a well known location to be automatically seeded. 

2. Dialog seeding setting the `system` flag for all dialogs:
 a. Creates a `system` Service dialog for all Service dialog(s) in that location.

I would caution against marking all of the seeded Service dialogs `system` unless it's absolutely necessary.

bdunne: My understanding of the `system` flag is that it would be used to determine which dialogs are provided by our plugins via seeding (always `system => true`) and which dialogs are user provided either imported from file or created in the UI (always `system => false`).  

This would allow us to:
  - delete any dialogs that were previously seeded and have since been removed from the plugin
  - avoid exporting seeded dialogs
  - prevent users from modifying seeded dialogs that could be overwritten by future changes to the plugin

I'm suggesting that the `seed` method should set `system => true` so that plugin authors don't need to remember it.  Forgetting to do so would allow users to change dialogs and those changes would get overwritten during the next seed (next server startup).  Users would probably open bugs about that.

> Gives providers the ability to provide OOTB non-system Service dialogs

Is that a thing we want/need?  Won't any user changes get overwritten by seed during the next server startup?

>  Gives users the ability to copy their own Service dialogs to a well known location to be automatically seeded.

Seed only looks at plugins, isn't this the purpose of Dialog Import / Export?

tinaafitz: Thanks @bdunne and @jrafanie, I guess I was thinking about the scenario where 'extra' dialogs copied to the plugin location would get loaded as `system` dialogs.  Is that a concern here?

jrafanie: @tinaafitz I'm not sure what the 'extra' dialogs are so I'm going to type some thoughts about `system => true`, let me know if anything is relevant to these extra dialogs. 

My understanding is anything we want to ship, maintain, and be able to update without worry about modifying customizations should be marked as system or read only.  If users are responsible for updating things, then these things shouldn't be system.  If we or a plugin who also wants to control the lifecycle of the "thing" (push changes without worry about user's customizing the base) creates a thing with that intent, we should mark it as system, so system as default seems like a good idea.

Perhaps, if users have a need to seed their customizations, they could opt-out of system => true but I don't know that's what is happening here.

Fryguy: Can we get some clarity on what the system column is for dialogs?  In other tables, it basically means "OOTB", which in turn implies that a) it is seeded and b) it's immutable by end users.

bdunne: Where are the 'extra' dialogs coming from?

If they're committed to the plugin repo or provided by our RPMs then I would argue that they are system dialogs.

If someone is manually copying them into the plugin directory, that's not a supportable thing.  Upgrading the RPMs will likely change the plugin directory since the git-based gem directory names include the git SHA.  `/opt/manageiq/manageiq-gemset/bundler/gems/manageiq-providers-ovirt-2074d5955a00/content/service_dialogs/transform-vm.yml` for exmple.  They should use dialog import instead.

bdunne: > Can we get some clarity on what the `system` column is for dialogs? In other tables, it basically means "OOTB", which in turn implies that a) it is seeded and b) it's immutable by end users.

https://github.com/ManageIQ/manageiq-schema/pull/468

d-m-u: I'm merely trying to close (a months old issue assigned to me:) https://github.com/ManageIQ/manageiq/issues/15144 that at this point is moot since we've largely gotten rid of the dialogs in question.

PR #512, opened 2020-07-23
MiqHashStruct was replaced with OpenStruct in provisioning

agrare:
In core the use of MiqHashStruct was replaced with OpenStruct in miq_provision_virt_workflow so we need to change the expectations here to match.

Test failures: https://travis-ci.com/github/ManageIQ/manageiq-providers-ovirt/jobs/364034924#L2046-L2060

Introduced by: https://github.com/ManageIQ/manageiq/pull/20374


agrare:
cc @gtanzillo

NickLaMuro:
crap!  Sorry @agrare for making more work for you 😭

agrare:
No worries it was a good fix so worth a small breakage here :smile:

PR #507, opened 2020-07-02
Add applications requests to VCRs

gekorob:
For each vm there will be a request to get applications.
This PR add mock requests to VCRs, returning empty applications section, as a temporary fix to specs failing

PR #519, opened 2020-08-31
[JANSA] Fix network spec VCR

gekorob:
Add applications GET request to old VCR

Related to: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/507#issuecomment-681922222


gekorob:
@agrare @simaishi This PR adds the missing call to VCR. I know that is not the best way, but in the master we are using different recordings and also expectations in the relative spec, that were changed by #482 (that was not back-ported as far as I know).
It seems to me the fast way to fix the error in Jansa

PR #529, opened 2020-10-06
Hosts should be disconnected not deleted

agrare:
Hosts are supposed to be disconnected so that they can be reconnected in the future.  https://github.com/ManageIQ/manageiq/pull/20588 fixed this in core and this test should be updated to match

Fixes https://travis-ci.com/github/ManageIQ/manageiq-providers-ovirt/jobs/396112357#L2106-L2115


mwperina:
Piotr, Moti or Boris could you please take a look?

@pkliczewski @masayag @borod108

PR #541, opened 2020-12-17
Fix linter warnings

djberg96:
This PR addresses most of the rubocop linter warnings. Specifically:

* ~~Disable the Lint/RescueException warning since we can't risk the breakage.~~
* Change what could be confused for conditionals into clearer code.
* Replace the long deprecated `URI.decode` with `URI.decode_www_form_component`
* Mark an unused variable in the `vm_hardware_guest_devices` method.

The one thing I couldn't fix was the call to `URI.escape` in the vm_import.rb file because of the second argument. I haven't found a replacement for it, except in the `Addressable` library, but that would mean adding a dependency.

### DIFF
--- a/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
@@ -390,7 +390,7 @@ class ManageIQ::Providers::Redhat::Inventory::Parser::InfraManager < ManageIQ::P
         :uid_ems          => vm.id,
         :connection_state => "connected",
         :vendor           => "redhat",
-        :name             => URI.decode(vm.name),
+        :name             => URI.decode_www_form_component(vm.name),
         :location         => "#{vm.id}.ovf",
         :template         => template,
         :memory_limit     => extract_vm_memory_policy(vm, :max),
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb
```diff
@@ -521,7 +521,7 @@ def hardware_disks(persister_hardware, disks)
     end
   end
 
-  def vm_hardware_guest_devices(persister_hardware, vm, addresses, host)
+  def vm_hardware_guest_devices(persister_hardware, vm, addresses, _host)
```
agrare, line 524: This is an internal method so we can just drop the `host` argument if it isn't used

djberg96: I'm always paranoid about interface compliance, but ok.

agrare:
```
../manageiq-providers-ovirt/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:    vm_hardware_guest_devices(persister_hardware, vm, addresses, host) unless template
../manageiq-providers-ovirt/app/models/manageiq/providers/redhat/inventory/parser/infra_manager.rb:  def vm_hardware_guest_devices(persister_hardware, vm, addresses, host)
```

djberg96:
I think I'm gonna leave this for others.

PR #543, opened 2021-01-13
Add a VmScan spec for Ovirt

agrare:
It was found that RHV SSA was broken due to missing state transitions (see https://github.com/ManageIQ/manageiq/pull/20953 for more info).  We noticed that there wasn't a vm_scan_spec.rb like the other providers that implement SSA.

This doesn't cover everything but is basically a copy of the specs already in vmware and openstack repos updated for Ovirt.


agrare:
This is mostly a copy of the vmware spec updated to work on Ovirt, but it still passes without the change to core to add the `start_scan` method https://github.com/ManageIQ/manageiq/pull/20953/files so it will require some changes to cover the bug

Fryguy (approved):

PR #559, opened 2021-05-06
Fix OVN NetworkManager subclass spec failures

agrare:
Fixes https://travis-ci.com/github/ManageIQ/manageiq-providers-ovirt/jobs/503285272#L2196-L2356

Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-openstack/pull/714

### DIFF
--- a/app/models/manageiq/providers/redhat/network_manager/cloud_network/private.rb
+++ b/app/models/manageiq/providers/redhat/network_manager/cloud_network/private.rb
@@ -1,0 +1,5 @@
+class ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork::Private < ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork
+  def self.display_name(number = 1)
+    n_('Cloud Network (Ovirt)', 'Cloud Networks (Ovirt)', number)
+  end
+end
--- a/app/models/manageiq/providers/redhat/network_manager/cloud_network/public.rb
+++ b/app/models/manageiq/providers/redhat/network_manager/cloud_network/public.rb
@@ -1,0 +1,5 @@
+class ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork::Public < ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork
+  def self.display_name(number = 1)
+    n_('External Cloud Network (Ovirt)', 'External Cloud Networks (Ovirt)', number)
+  end
+end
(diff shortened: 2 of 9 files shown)


agrare:
Cross-repo tests are green: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/364

chessbyte:
is oVirt really using classes from Openstack?

agrare:
Yes, the OVN network manager is essentially an Openstack compatible neutron API so we re-use the openstack collector/parser/etc...

https://www.ovirt.org/develop/release-management/features/network/ovirt-ovn-provider.html
https://www.ovirt.org/develop/release-management/features/network/manageiq_ovn.html

app/models/manageiq/providers/redhat/network_manager/cloud_network/private.rb
```diff
@@ -0,0 +1,5 @@
+class ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork::Private < ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork
+  def self.display_name(number = 1)
+    n_('Cloud Network (Ovift)', 'Cloud Networks (Ovirt)', number)
```
Fryguy, line 3:
```suggestion
    n_('Cloud Network (Ovirt)', 'Cloud Networks (Ovirt)', number)
```

app/models/manageiq/providers/redhat/network_manager/cloud_network/public.rb
```diff
@@ -0,0 +1,5 @@
+class ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork::Public < ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork
```
Fryguy, line 1:
```suggestion
class ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork::Public < ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork
```

app/models/manageiq/providers/redhat/network_manager/cloud_network/private.rb
agrare: :facepalm: fixed

Fryguy:
@agrare Does this require a data migration?  It's hard to tell, but it looks like Public and Private are new classes.

agrare:
Actually those are existing Openstack classes but in order to have them show up as subclasses of `ManageIQ::Providers::Redhat::NetworkManager::CloudNetwork` they had to inherit from that rather than their Openstack equivalents.

And yes we need a data migration to update the type for any Redhat::NetworkManager

Fryguy:
`morphy` version: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/569

PR #569, opened 2021-10-06
[MORPHY] Fix Redhat::NetworkManager refresh STI failures

agrare:
The Openstack parser sets inventory to a hard-coded `:type` of `Openstack::NetworkManager::*` which breaks for the RHEV subclass because they don't aren't under the Redhat::NetworkManager namespace

Alternative to backporting https://github.com/ManageIQ/manageiq-providers-openstack/pull/714 and https://github.com/ManageIQ/manageiq-providers-ovirt/pull/559 due to the schema migration


app/models/manageiq/providers/redhat/network_manager.rb
```diff
@@ -1,14 +1,9 @@
 class ManageIQ::Providers::Redhat::NetworkManager < ManageIQ::Providers::NetworkManager
-  require_nested :CloudNetwork
-  require_nested :CloudSubnet
   require_nested :EventCatcher
   require_nested :EventParser
   require_nested :FloatingIp
```
agrare, line 4: I didn't have to delete this for the specs to pass, but I don't see why this wouldn't have the same issue.  Investigating...

agrare: Looks like there just weren't any floating_ips in this VCR, still need to delete this

Issue #526, opened 2020-09-28
Validating credentials when adding a provider always passes


skateman:
When trying to add a provider with invalid credentials (even a nonexisting host) it always passes...

@miq-bot add_label bug
@miq-bot assign @agrare

skateman:
cc @mzazrivec

mzazrivec:
@agrare this one can be closed.

mzazrivec:
Actually, I spoke too soon. I mistakenly thought https://github.com/ManageIQ/manageiq-providers-ovirt/pull/525 would fix this issue, but it did not. The problem described above is still present and needs to be addressed in backend.

zigaSRC:
@mzazrivec @agrare This bug is still present in lasker-1. 

Furthermore, authentication for metrics collection for ovirt provider is no longer working. Manageiq no longer collects the metrics in lasker-1. Collection/authentication worked in jansa-2. 

```
[----] W, [2021-09-07T09:52:32.674478 #2502:2ad348893960]  WARN -- evm: MIQ(ManageIQ::Providers::Redhat::InfraManager#authentication_che
ck_no_validation) type: [:metrics] for [6] [srcOLVM] Validation failed: invalid, undefined method `async_exec' for nil:NilClass
[----] W, [2021-09-07T09:52:32.674650 #2502:2ad348893960]  WARN -- evm: MIQ(AuthUseridPassword#validation_failed) [ExtManagementSystem]
[6], previously valid on: 2021-07-21 09:15:25 UTC, previous status: [Invalid]
```

Issue #580, opened 2021-12-01
Metrics auth validation fails with Validation failed: invalid, undefined method `async_exec' for nil:NilClass


zackary-deems:
Infrastructure Management 2.3.0 (MiQ Morphy) while setting up a RHV 4.4 provider, fails on validation of the metrics datastore credentials.  In the UI it appears to always return 'Valid', but in the evm.log it shows validation failed with the above... and does so as a WARN, not an error.

zackary-deems:
Attaching the last 5000 lines of the evm.log
[evm.sample.log](https://github.com/ManageIQ/manageiq-providers-ovirt/files/7636386/evm.sample.log)

agrare:
Thanks for the logs @zackary-deems, I don't see the exception that you note here in this log.  Can you check if it was maybe in the previous check?

zackary-deems:
[evm.sample.log](https://github.com/ManageIQ/manageiq-providers-ovirt/files/7636730/evm.sample.log)

agrare:
Thanks @zackary-deems it looks like the ovirt code is swallowing the backtrace, can you add the following code and re-run the verify task?

```
cd /var/www/miq/vmdb
bundle show manageiq-providers-ovirt
```

This will tell you what directory the ovirt provider is in, e.g. mine outputs `/opt/manageiq/manageiq-gemset/bundler/gems/manageiq-providers-ovirt-7f18ec90ba85` for a master appliance.

Then open this file `app/models/manageiq/providers/redhat/infra_manager/api_integration.rb` and add this line:
```
     def handle_credentials_verification_error(err)
+      _log.log_backtrace(err)
       case err
```

Once that is saved run `systemctl restart evmserverd` and re-try the verification.  We should get a backtrace in the evm.log after that.

agrare:
It sounds like a bug coming from ovirt_metrics since async_exec is a postgres gem method and ovirt_metrics does some weird active record connection adapter magic

agrare:
The `async_exec` backtrace is:
```
"/home/grare/adam/.gem/gems/activerecord-6.0.4.1/lib/active_record/connection_adapters/postgresql/database_statements.rb:92:in `block (2 levels) in execute'",
 "/home/grare/adam/.gem/gems/activesupport-6.0.4.1/lib/active_support/dependencies/interlock.rb:48:in `block in permit_concurrent_loads'",
[...]

PR #581, opened 2021-12-01
Fix metrics_endpoint info not passed to check_connect_metrics

agrare:
The `check_connect_metrics` method returns true if the server is nil, the expectation being that no metrics endpoint info was passed by the user.

The metrics endpoint keys in the options have changed since the DDF changes, which resulted in the metrics_server _always_ being nil. This meant that no matter what info was passed in the form it would verity successfully.

Fixes https://github.com/ManageIQ/manageiq-providers-ovirt/issues/526

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -224,8 +224,8 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       )
 
       metrics_username, metrics_password = metrics_authentication&.values_at("userid", "password")
-      metrics_port, metrics_database = metrics_endpoint&.values_at(
-        "metrics_port", "metrics_database"
+      metrics_server, metrics_port, metrics_database = metrics_endpoint&.values_at(
+        "hostname", "port", "path"
       )
 
       !!raw_connect(
@@ -237,6 +237,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
         :ca_certs         => ca_certs,
         :metrics_username => metrics_username,
         :metrics_password => ManageIQ::Password.try_decrypt(metrics_password),
+        :metrics_server   => metrics_server,
         :metrics_port     => metrics_port,
         :metrics_database => metrics_database
       )


app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -224,8 +225,8 @@ def verify_credentials(args)
       )
 
       metrics_username, metrics_password = metrics_authentication&.values_at("userid", "password")
-      metrics_port, metrics_database = metrics_endpoint&.values_at(
-        "metrics_port", "metrics_database"
```
agrare, line 228: After the DDF form changes the `args["endpoints"]["metrics"]` hash looks like `{"hostname"=>"localhost", "port"=>5432, "path"=>"ovirt_engine_history"}` we no longer prefix with the endpoint name because we aren't mashing all keys into one flat hash anymore.

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -237,6 +238,7 @@ def verify_credentials(args)
         :ca_certs         => ca_certs,
         :metrics_username => metrics_username,
         :metrics_password => ManageIQ::Password.try_decrypt(metrics_password),
+        :metrics_server   => metrics_server,
```
agrare, line 241: We also never even passed the metrics_server in, so honestly not sure how this ever worked

app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -169,6 +169,7 @@ def rethrow_as_a_miq_error(ovirt_sdk_4_error)
     end
 
     def handle_credentials_verification_error(err)
+      _log.log_backtrace(err)
```
Fryguy, line 172: This is debug, right?

agrare: `log_backtrace` is `WARN` I believe

Fryguy: oh I didn't recognize that this method was precisely for errors

agrare: Hm something doesn't seem right about this though:
`[----] E, [2021-12-01T16:57:07.158658 #1699986:91f0] ERROR -- evm: [MiqException::MiqEVMLoginError]: undefined method `async_exec' for nil:NilClass  Method:[block (2 levels) in <class:LogProxy>]`

That's not a backtrace...

agrare: I'm going to take this out since it isn't directly related to the check_connect_metrics fix and I need to see why it isn't logging the backtrace properly.

Fryguy (approved):

agrare:
I'm going to take `Fixes #580` off of this because that mentions the failing `exec_async` where this fixes the issue that everything always suceeds.

Fryguy:
Backported to `morphy` in commit e1f06c723cfbf4accf633a105e46d1f21729aedd.

```text
commit e1f06c723cfbf4accf633a105e46d1f21729aedd
Date:   Fri Dec 3 11:20:12 2021 -0500

    Merge pull request #581 from agrare/fix_metrics_server_not_passed
    
    Fix metrics_endpoint info not passed to check_connect_metrics
    
    (cherry picked from commit 5c817eedecd6aa11abe2ea8cd8e13acd7ab2fa52)
```

Fryguy:
Backported to `lasker` in commit c48c8735e7be1ad133160b413b4361349a842f1e.

```text
commit c48c8735e7be1ad133160b413b4361349a842f1e
Date:   Fri Dec 3 11:20:12 2021 -0500

    Merge pull request #581 from agrare/fix_metrics_server_not_passed
    
    Fix metrics_endpoint info not passed to check_connect_metrics
    
    (cherry picked from commit 5c817eedecd6aa11abe2ea8cd8e13acd7ab2fa52)
```

PR #582, opened 2021-12-01
Fix default api verification always succeeding

agrare:
The `verify_credentials` was assuming the `verify_ssl` option was a boolean not an integer (also updated with the DDF changes), which meant that `verify_ssl ? VERIFY_PEER : VERIFY_NONE` was always returning VERIFY_PEER even if you wanted VERIFY_NONE.

Which didn't matter anyway because the exception handler would only raise if the exception string matched `/error.*sso/i` which it wouldn't if you had anything other than a username/password exception.  This meant you could point the RHV provider to literally anything and it would verify successfully :scream:

### DIFF
--- a/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
+++ b/app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
@@ -233,7 +233,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
         :password         => ManageIQ::Password.try_decrypt(password),
         :server           => server,
         :port             => port,
-        :verify_ssl       => verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE,
+        :verify_ssl       => verify_ssl,
         :ca_certs         => ca_certs,
         :metrics_username => metrics_username,
         :metrics_password => ManageIQ::Password.try_decrypt(metrics_password),
@@ -321,9 +321,7 @@ module ManageIQ::Providers::Redhat::InfraManager::ApiIntegration
       begin
         connection = raw_connect_v4(opts)
         connection.test(:raise_exception => true)
-        return true
-      rescue OvirtSDK4::Error => error
-        raise error if /error.*sso/i.match?(error.message)
+        true
       end
     end
 


app/models/manageiq/providers/redhat/infra_manager/api_integration.rb
```diff
@@ -233,7 +233,7 @@ def verify_credentials(args)
         :password         => ManageIQ::Password.try_decrypt(password),
         :server           => server,
         :port             => port,
-        :verify_ssl       => verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE,
+        :verify_ssl       => verify_ssl,
```
Fryguy, line 236: I think this interface used to take either an integer (that is an OpenSSL::VERIFY* constant) or a boolean.  Do we still want the backend verify method to support both or do we only want to support integers?

kbrock: also, the ruby internal code handles nil in a strange way in terms of verifying the peer.

agrare: Maybe?  Do you mean this code in particular or `self.verify_credentials` in general

This code looks like if verify_ssl is an int it will always `VERIFY_PEER`

agrare: > also, the ruby internal code handles nil in a strange way in terms of verifying the peer.

I can `verify_ssl ||=` to ensure this is never nil if that'd help, these args come from the DDF params_for_create which has a default value and so this shouldn't ever be nil (famous last words of course)

Fryguy: I meant verify_credentials in general.  I thought we had sort of casually of agreed on supporting both a boolean (for convenience, e.g. in rails console) and a integer (for accuracy, e.g. via DDF).  I'm fine with any, just thought it made more sense to decide something now and then ultimately ensure the interface is consistent.

Fryguy: > This code looks like if verify_ssl is an int it will always VERIFY_PEER

Yeah the original code is written as the boolean case, and apparently didn't ever support integers so it is wrong regardless.  I'm more trying to understand what is interface we _want_ to support.  I had thought we wanted both boolean and integer, which is why this changed jumped out at me, cause I was expecting something roughly more like:

```ruby
        verify_ssl = verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE unless verify_ssl.is_a?(Integer)
        :verify_ssl  => verify_ssl
```

Fryguy: Actually, looking higher up the method, it seems this has always come out of the default_endpoint, and will always be an integer.  I didn't realize this particular method `verify_credentials` was reworked to be _only_ DDF based (was surprised to see the args don't allow for overwriting the value, but the args has completely changed). Kind of wish this method was named verify_credentials_ddf :(

Sorry for the tangent, but this is fine.

Fryguy:
Backported to `morphy` in commit 26f11891fad76dbb4b215d24cc6c024e5e062049.

```text
commit 26f11891fad76dbb4b215d24cc6c024e5e062049
Date:   Fri Dec 3 11:05:56 2021 -0500

    Merge pull request #582 from agrare/fix_default_verify_ssl_and_rescue
    
    Fix default api verification always succeeding
    
    (cherry picked from commit 204e7e4e3654ea5997b63d40e491f68c7ef5c2b6)
```

Fryguy:
Backported to `lasker` in commit 5bf1523f48b9aafdeaa5e68164d90b3a233fefbd.

```text
commit 5bf1523f48b9aafdeaa5e68164d90b3a233fefbd
Date:   Fri Dec 3 11:05:56 2021 -0500

    Merge pull request #582 from agrare/fix_default_verify_ssl_and_rescue
    
    Fix default api verification always succeeding
    
    (cherry picked from commit 204e7e4e3654ea5997b63d40e491f68c7ef5c2b6)
```

PR #604, opened 2022-05-31
Add EmsRefresherHelper to refresher spec

nasark:
- due to recent changes of adding a `last_refresh_success_date` column, this test has been switched to use `EmsRefresherHelper` so that we can use the `assert_inventory_not_changed` helper method and skip checks for the `last_refresh_success_date` attribute (https://github.com/ManageIQ/manageiq/pull/21889)

@miq-bot add_label test
@miq-bot assign @agrare


spec/models/manageiq/providers/redhat/infra_manager/refresher/refresher_graph_target_host_spec.rb
```diff
@@ -16,18 +17,16 @@
                        .to receive(:collect_vnic_profiles).and_return([])
     end
 
-    let(:models_for_host_target) { [ExtManagementSystem, EmsFolder, EmsCluster, Storage, HostStorage, Switch, HostSwitch, Lan, CustomAttribute] }
-
     it 'does not change the host when target refresh after full refresh' do
       EmsRefresh.refresh(@ems)
       @ems.reload
 
-      saved_inventory = serialize_inventory(models_for_host_target)
```
agrare, line 25: We should be able to delete this old method from https://github.com/nasark/manageiq-providers-ovirt/blob/8ecdbac03210ff73ed9c4e014fc037ad2b5720b5/spec/models/manageiq/providers/redhat/infra_manager/refresher/ovirt_refresher_spec_common.rb#L6-L15 now too

agrare (approved):
Awesome

PR #605, opened 2022-06-07
[WIP] Don't delegate NetworkManager#authentication_check

agrare:
The Ovirt OVN NetworkManager uses OpenStack APIs to connect, therefore we shouldn't simply use the ovirt/rhv API auth check to indicate if we are able to connect to an OVN network manager.


agrare:
This started as a result of looking at a RHV NetworkManager EventCatcher spinning at 100% cpu, still investigating

PR #620, opened 2022-11-09
Add support for creating cloud network/subnets/routers

agrare:
Add Ovirt OVN/NetworkManager support for creating cloud networks, subnets, and routers.

The OVN API is an Openstack Neutron compatible API so we can re-use the CRUD methods from the Openstack subclass.

Fixes https://github.com/ManageIQ/manageiq/issues/22109#issuecomment-1262262868

Screenshot from 2022-11-09 16-46-36

Screenshot from 2022-11-09 16-46-55

Screenshot from 2022-11-09 16-47-12

Screenshot from 2022-11-09 16-48-13

### DIFF
--- a/app/models/manageiq/providers/ovirt/network_manager.rb
+++ b/app/models/manageiq/providers/ovirt/network_manager.rb
@@ -13,6 +13,9 @@ class ManageIQ::Providers::Ovirt::NetworkManager < ManageIQ::Providers::NetworkM
   include ManageIQ::Providers::Openstack::ManagerMixin
   include SupportsFeatureMixin
 
+  supports :create_network_router
+  supports :cloud_subnet_create
+
   has_many :public_networks,  :foreign_key => :ems_id, :dependent => :destroy,
            :class_name => "ManageIQ::Providers::Openstack::NetworkManager::CloudNetwork::Public"
   has_many :private_networks, :foreign_key => :ems_id, :dependent => :destroy,


app/models/manageiq/providers/ovirt/network_manager.rb
```diff
@@ -13,6 +13,9 @@ class ManageIQ::Providers::Ovirt::NetworkManager < ManageIQ::Providers::NetworkM
   include ManageIQ::Providers::Openstack::ManagerMixin
   include SupportsFeatureMixin
 
+  supports :create_network_router
+  supports :cloud_subnet_create
```
agrare, line 17: NOTE It appears that the UI uses this feature to test for support of all of the network create features, there is no `create_cloud_network` feature.

IMO we should:
1. Add a `create_cloud_network` feature to core and all supporting providers
2. Update the UI to check for the right feature for the action being taken
3. Standardize on `create_cloud_subnet` since it is reversed compared with `create_network_router`

kbrock: can we use `CloudNetwork.supports :create` instead of putting that here?

agrare: Yeah definitely, you mean like https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L303 (with `:child_model`)

agrare: @kbrock do you want to do that in this PR?  I was thinking we would add the functionality the way it is currently implemented in this provider, then follow-up with refactoring

Fryguy:
Please also add to the capabilities matrix.

PR #621, opened 2022-11-17
Subclass IsoDatastore and transfer from core

nasark:
- since iso datastores are only used by oVirt/RHV, we can move its model from core and into the provider plugin
- iso datastores will be modelled as `Storage` objects with STI `:type => ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore`
- `synchronize_advertised_images` functionality is now in a refresh and the images will be persisted

Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/22239
- [ ] https://github.com/ManageIQ/manageiq-schema/pull/672
- [ ] https://github.com/ManageIQ/manageiq-api/pull/1186
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/8548
- [ ] https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/17
- [x] https://github.com/ManageIQ/manageiq/pull/22258

TODO:
- [X] `IsoDatastore` references in UI
- [x] supports :iso_datastore feature
- [x] Test with `manageiq-providers-red_hat_virtualization`

Ref:
- https://github.com/ManageIQ/manageiq/issues/19440

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/ovirt/infra_manager.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager.rb
@@ -17,6 +17,7 @@ class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManag
   require_nested  :ProvisionWorkflow
   require_nested  :Snapshot
   require_nested  :Storage
+  require_nested  :IsoDatastore
   require_nested  :Template
   require_nested  :Vm
   require_nested  :DistributedVirtualSwitch
@@ -27,6 +28,8 @@ class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManag
   has_many :vm_and_template_ems_custom_fields, :through => :vms_and_templates, :source => :ems_custom_attributes
   has_many :external_distributed_virtual_switches, :dependent => :destroy, :foreign_key => :ems_id, :inverse_of => :ext_management_system
   has_many :external_distributed_virtual_lans, -> { distinct }, :through => :external_distributed_virtual_switches, :source => :lans
+  has_many :iso_datastores, :dependent => :destroy, :foreign_key => :ems_id, :inverse_of => :ext_management_system
+  has_many :iso_images, :through => :storages
 
   include HasNetworkManagerMixin
 
@@ -42,7 +45,7 @@ class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManag
   end
 
   supports :create_iso_datastore do
-    unsupported_reason_add(:create_iso_datastore, _("Already has an ISO datastore")) if iso_datastore
+    unsupported_reason_add(:create_iso_datastore, _("Already has an ISO datastore")) unless iso_datastores.empty?
   end
 
   def ensure_managers
--- a/app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
@@ -1,0 +1,7 @@
+class ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore < ManageIQ::Providers::Ovirt::InfraManager::Storage
+  belongs_to :ext_management_system, :foreign_key => :ems_id
+
+  def self.display_name(number = 1)
+    n_('ISO Datastore', 'ISO Datastores', number)
+  end
+end
--- a/app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
@@ -1,4 +1,6 @@
 class ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore < ManageIQ::Providers::Ovirt::InfraManager::Storage
+  supports :iso_datastore
+
   belongs_to :ext_management_system, :foreign_key => :ems_id
 
   def self.display_name(number = 1)
--- a/app/models/manageiq/providers/ovirt/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/ovirt/inventory/collector/target_collection.rb
@@ -150,6 +150,10 @@ class ManageIQ::Providers::Ovirt::Inventory::Collector::TargetCollection < Manag
     t
   end
 
+  def advertised_images
+    []
+  end
+
   def parse_targets!
     target.targets.each do |t|
       case t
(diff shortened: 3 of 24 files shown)


nasark:
Update: `supports :iso_datastore` feature can be done in a follow-up

app/models/manageiq/providers/ovirt/infra_manager.rb
```diff
@@ -42,7 +44,7 @@ class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManag
   end
 
   supports :create_iso_datastore do
-    unsupported_reason_add(:create_iso_datastore, _("Already has an ISO datastore")) if iso_datastore
+    unsupported_reason_add(:create_iso_datastore, _("Already has an ISO datastore")) if iso_datastores
```
agrare, line 47: I think you want `if iso_datastores.any?` or `unless iso_datastores.empty?` since `[]` is going to return truthy

app/models/manageiq/providers/ovirt/inventory/parser/infra_manager.rb
```diff
@@ -113,6 +114,19 @@ def storagedomains
     end
   end
 
+  def advertised_images
+    collector.advertised_images.each do |image|
```
agrare, line 118: Are the advertised images an ISO datastore?  I thought they would be IsoImages that belong to an IsoDatastore

app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
```diff
@@ -0,0 +1,54 @@
...
+  def synchronize_advertised_images_queue
+    MiqQueue.put_unless_exists(
+      :class_name  => self.class.name,
+      :instance_id => id,
+      :method_name => "synchronize_advertised_images",
+      :queue_name  => ext_management_system.queue_name_for_ems_operations,
+      :zone        => ext_management_system.try(:my_zone),
+      :role        => "ems_operations"
+    )
+  end
```
agrare, line 20: If we're getting these in refresh below do we need this still?

app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    db_image_hash.each_value(&:destroy)
+
+    clear_association_cache
+    update_attribute(:last_refresh_on, Time.now.utc)
+
+    _log.info("Synchronizing images on #{log_for}...Complete")
+  rescue ManageIQ::Providers::Ovirt::InfraManager::OvirtServices::Error
+  end
```
agrare, line 49: Same comment here actually, if we move the `iso_images` mapping into refresh this can all go away

app/models/manageiq/providers/ovirt/inventory/parser/infra_manager.rb
nasark: Yes you are correct they belong to an IsoDatastore. I had a brain fart moment here haha

app/models/manageiq/providers/ovirt/infra_manager/iso_datastore.rb
nasark: No these are no longer needed, thanks

nasark:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22239, https://github.com/ManageIQ/manageiq-ui-classic/pull/8548, https://github.com/ManageIQ/manageiq-api/pull/1186, https://github.com/ManageIQ/manageiq-schema/pull/672, https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/17

spec/factories/iso_datastore.rb
```diff
@@ -0,0 +1,3 @@
+FactoryBot.define do
+  factory :iso_datastore, :class => 'ManageIQ::Providers::Ovirt::InfraManager::IsoDatastore'
```
agrare, line 2: Think we should have `, :parent => :storage_redhat`

app/models/manageiq/providers/ovirt/inventory/collector/target_collection.rb
```diff
@@ -150,6 +150,10 @@ def templates
     t
   end
 
+  def advertised_images
+    []
```
agrare, line 154: So we do handle targeted refresh of storage domains, but the `manager.ovirt_services.advertised_images` call only works for all storagedomains.

A good follow-up would be to extract the `advertised_images` method into the collector, then adding a targeted advertised_images that looks at `references(:storagedomains)`

app/models/manageiq/providers/ovirt/inventory/parser/infra_manager.rb
```diff
@@ -98,17 +99,32 @@ def storagedomains
       total       = free + used
       committed   = storagedomain.try(:committed).to_i
 
+      storage_domain_type = storagedomain.dig(:type, :downcase)
```
agrare, line 102: This looks weird, is there a `:downcase` key in the `:type` hash?  I would have expected `storagedomain.dig(:type)&.downcase`

nasark: I think I got it from the old code https://github.com/ManageIQ/manageiq-providers-ovirt/pull/621/files/8fe7666ebdd56608e2858ea353e21ad0372a3f82#diff-43e38e24277dd68083169041f77e7830443aab1372bb4e7c472e3ad596cefffcL105, but I think `storagedomain.dig(:type)&.downcase` would be better?

agrare: ah yeah you're right, so `storagedomain` isn't a hash, so this isn't the standard dig
```
(byebug) storagedomain.method(:dig)
#<Method: OvirtSDK4::StorageDomain(OvirtSDK4::Type)#dig(*keys) /home/grare/adam/.gem/ruby/3.0.0/gems/ovirt-engine-sdk-4.4.1/lib/ovirtsdk4/type.rb:79>
```

So this does work:
```
(byebug) storagedomain.dig(:type)
"iso"
(byebug) storagedomain.dig(:type, :downcase)
"iso"
```

Carry on

agrare (approved):

nasark:
Cross-repo tests are [green](https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/724) (`manageiq-api` failures are unrelated)

PR #623, opened 2022-12-09
Fix RHV `#find_destination_in_vmdb` looking for the wrong type

agrare:
Fix RHV provisioning when calling `#find_destination_in_vmdb` due to looking for the wrong class name.

### DIFF
--- a/app/models/manageiq/providers/ovirt/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager/provision/cloning.rb
@@ -10,11 +10,8 @@ module ManageIQ::Providers::Ovirt::InfraManager::Provision::Cloning
   end
 
   def find_destination_in_vmdb(ems_ref)
-    if source.template?
-      ManageIQ::Providers::Ovirt::InfraManager::Vm.find_by(:name => dest_name, :ems_ref => ems_ref)
-    else
-      ManageIQ::Providers::Ovirt::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
-    end
+    assoc = source.template? ? ems.vms : ems.miq_templates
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ovirt/infra_manager/provision/cloning.rb
```diff
@@ -10,11 +10,8 @@ def destination_image_locked?
   end
 
   def find_destination_in_vmdb(ems_ref)
-    if source.template?
-      ManageIQ::Providers::Ovirt::InfraManager::Vm.find_by(:name => dest_name, :ems_ref => ems_ref)
-    else
-      ManageIQ::Providers::Ovirt::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
-    end
```
agrare, line 17: The old way did not include the ems_id in the query and could identify a VM on another EMS :cry:

app/models/manageiq/providers/ovirt/infra_manager/provision/cloning.rb
```diff
@@ -10,11 +10,8 @@ def destination_image_locked?
...
+    assoc = source.template? ? :vms : :miq_templates
+    ems.send(assoc).find_by(:name => dest_name, :ems_ref => ems_ref)
```
Fryguy, line 14: Super-minor, but we can avoid the `.send`, if you think that's nicer - up to you:

```suggestion
    assoc = source.template? ? ems.vms : ems.miq_templates
    assoc.find_by(:name => dest_name, :ems_ref => ems_ref)
```

agrare: Yeah that's nicer

kbrock (approved):
lgtm

PR #625, opened 2022-12-15
Simplify provider subclasses of Ovirt

agrare:
There are a number of classes and/or methods which have to be copied completely to work properly due to hard coded references to Ovirt::InfraManager / "ovirt"

Dependents:
* https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/18


agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/18

agrare:
There is still more to go, but I don't want these PRs to be enormous.

Taking out of WIP

app/models/manageiq/providers/ovirt/infra_manager.rb
```diff
@@ -112,6 +117,14 @@ def self.description
     @description ||= "oVirt".freeze
   end
 
+  def self.vm_vendor
+    "ovirt".freeze
```
Fryguy, line 121: Super minor but should these be memoized like the ones above?

Fryguy:
One minor comment, but merging anyway

Issue #631, opened 2023-02-17
OVN network_manager event catcher always fails


agrare:
The Ovirt OVN service doesn't not have any service which can be used for events leaving the event_catcher to fail continuously:

```
[----] I, [2023-02-16T14:21:00.694829 #376658:a1f4]  INFO -- evm: ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner started. ID [350], PID [376658], GUID [f7913259-7f5f-4bba-9429-21c97bfc68c2], Zone [default], Role [automate,database_operations,database_owner,ems_inventory,event,reporting,scheduler,user_interface,web_services]
[----] I, [2023-02-16T14:21:00.694922 #376658:a1f4]  INFO -- evm: MIQ(ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [RHVdev1.rtp.raleigh.ibm.com] as [admin@internal] Validating Connection/Credentials
[----] I, [2023-02-16T14:21:01.121992 #376658:a1f4]  INFO -- evm: MIQ(ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [RHVdev1.rtp.raleigh.ibm.com] as [admin@internal] Starting Event Monitor Thread
[----] I, [2023-02-16T14:21:01.122281 #376658:a1f4]  INFO -- evm: MIQ(ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [RHVdev1.rtp.raleigh.ibm.com] as [admin@internal] Started Event Monitor Thread
[----] I, [2023-02-16T14:21:16.266152 #376658:a1f4]  INFO -- evm: MIQ(ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner#do_work) EMS [RHVdev1.rtp.raleigh.ibm.com] as [admin@internal] Event Monitor Thread gone. Restarting...
[----] I, [2023-02-16T14:21:16.266621 #376658:a1f4]  INFO -- evm: MIQ(ManageIQ::Providers::Redhat::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [RHVdev1.rtp.raleigh.ibm.com] as [admin@internal] Validating Connection/Credentials
[...]

PR #632, opened 2023-02-17
Remove the OVN NetworkManager Event Catcher

agrare:
RHV with the OVN service enabled does not have ceilometer, stf, or amqp so there is nothing for the event catcher to connect to.

Schema migration: https://github.com/ManageIQ/manageiq-schema/pull/684

https://github.com/ManageIQ/manageiq-providers-ovirt/issues/631


Fryguy:
@agrare Not sure why this is failing as the error seems unrelated.

agrare:
Oh it is because there was a `{:class_name => "ManageIQ::Providers::Ovirt::NetworkManager", :method_name => "stop_event_monitor"}` on the queue due to the authentication change, and now that isn't there anymore.  Will update the count

agrare:
@Fryguy fixed

config/settings.yml
```diff
@@ -64,17 +64,6 @@
     :event_catcher:
       :event_catcher_ovirt:
         :poll: 15.seconds
-      :event_catcher_ovirt_network:
```
Fryguy, line 67: Do we need a data migration to remove these from the settings changes table if they exist?

agrare: :+1: yeah that is true I'll update the schema PR with this

agrare:
Will need to create a petrosian direct PR similar to https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/25

PR #635, opened 2023-03-08
convert console_supported? to supports?

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/21990

note, there is some duplicate code with https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/24

let me know how you want to handle this


kbrock:
update:
- fixed cops

PR #637, opened 2023-03-21
Fix VM Reconfigure for stopped VMs

agrare:
Ovirt VMs have no host assigned when they are not running.  This causes an exception to be raised when trying to reconfigure a VM that isn't running because we are using the host for the list of available lans.

### DIFF
--- a/spec/models/manageiq/providers/ovirt/infra_manager/vm/reconfigure_spec.rb
+++ b/spec/models/manageiq/providers/ovirt/infra_manager/vm/reconfigure_spec.rb
@@ -373,6 +373,7 @@ describe ManageIQ::Providers::Ovirt::InfraManager::Vm::Reconfigure do
   describe 'available_vlans' do
     let(:dc1) { FactoryBot.create(:datacenter_ovirt, :name => 'dc1', :ems_ref => 'dc1-ems-ref', :ems_ref_type => 'Datacenter') }
     let(:cluster) { FactoryBot.create(:ems_cluster, :uid_ems => "uid_ems", :name => 'cluster') }
+    let(:rp) { FactoryBot.create(:resource_pool, :ext_management_system => ems) }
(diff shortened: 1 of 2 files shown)


agrare:
Keeping this WIP because I think we can still return a list of Lans without a host we just need to looks for dvswitches in the vm's parent_datacenter.

spec/models/manageiq/providers/ovirt/infra_manager/vm/reconfigure_spec.rb
```diff
@@ -421,6 +421,14 @@
           expect(vlans).to match_array([lan_mgmt.name, lan_A.name, "#{ext_lan.name}/#{ext_dist_switch.name}"])
         end
 
+        context 'with a powered off vm' do
+          let(:vm) { FactoryBot.create(:vm_ovirt, :ext_management_system => ems, :storage => storage, :raw_power_state => 'down') }
+
+          it 'returns only external lans' do
```
agrare, line 427: I want to test the `parent_cluster` path but will need to set up the whole ems_metadata relationships tree for that to work

agrare: Forgot how easy it was with the `add_child` method, done

Fryguy (approved):

Fryguy:
Backported to `petrosian` in commit 312320ba13573d374731b6fedb6bc434b64e5f68.

```text
commit 312320ba13573d374731b6fedb6bc434b64e5f68
Date:   Tue Mar 21 16:39:24 2023 -0400

    Merge pull request #637 from agrare/fix_vm_reconfig_nil_host
    
    Fix VM Reconfigure for stopped VMs
    
    (cherry picked from commit adbe19ce083dbd24c3d05b507dbf33d8c32f7d95)
```

PR #653, opened 2023-10-11
Check supported storage type for vm Smartstate analysis scan

nasark:
Checking if vm storage type is supported for smartstate analysis scans when checking if vm supports smartstate analysis

@miq-bot add_label bug
@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/ovirt/infra_manager/vm_or_template_shared/scanning.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager/vm_or_template_shared/scanning.rb
@@ -7,6 +7,11 @@ module ManageIQ::Providers::Ovirt::InfraManager::VmOrTemplateShared::Scanning
       unless feature_supported
         unsupported_reason_add(:smartstate_analysis, reason)
       end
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ovirt/infra_manager/vm_or_template_shared/scanning.rb
```diff
@@ -7,6 +7,9 @@ module ManageIQ::Providers::Ovirt::InfraManager::VmOrTemplateShared::Scanning
       unless feature_supported
         unsupported_reason_add(:smartstate_analysis, reason)
       end
```
agrare, line 9: Should we add the `storage.nil?` check here?  https://github.com/ManageIQ/manageiq/pull/22738/files#diff-2dfd7951e9d3cde993582a8eb3f50a18c8728bf88db6216424e6b1bd6d1fabfaL250

nasark:
@miq-bot add_label quinteros/yes?

agrare (approved):

PR #656, opened 2024-01-26
Drop require_nested and include_concern

jrafanie:
Zeitwerk completely removes the need for these methods. See also: https://github.com/ManageIQ/manageiq/pull/22854

### DIFF
--- a/app/models/manageiq/providers/ovirt/inventory/collector.rb
+++ b/app/models/manageiq/providers/ovirt/inventory/collector.rb
@@ -1,8 +1,5 @@
 class ManageIQ::Providers::Ovirt::Inventory::Collector < ManageIQ::Providers::Inventory::Collector
   # TODO: review the changes here and find common parts with ManageIQ::Providers::Ovirt::InfraManager::Inventory::Strategies::V4
-  require_nested :InfraManager
-  require_nested :TargetCollection
-
   attr_reader :clusters
   attr_reader :networks
   attr_reader :storagedomains
--- a/app/models/manageiq/providers/ovirt/infra_manager.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager.rb
@@ -1,6 +1,6 @@
 class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManager
-  include_concern :ApiIntegration
-  include_concern :AdminUI
+  include ApiIntegration
+  include AdminUI
 
   has_many :cloud_tenants, :foreign_key => :ems_id, :dependent => :destroy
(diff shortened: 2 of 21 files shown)


app/models/manageiq/providers/ovirt/inventory/collector.rb
```diff
@@ -1,8 +1,5 @@
 class ManageIQ::Providers::Ovirt::Inventory::Collector < ManageIQ::Providers::Inventory::Collector
   # TODO: review the changes here and find common parts with ManageIQ::Providers::Ovirt::InfraManager::Inventory::Strategies::V4
```
Fryguy, line 2: This comment was attached to those require lines... Not sure if it should move somewhere else

agrare:
@jrafanie do these PRs have any dependencies?  On https://github.com/ManageIQ/manageiq/pull/22854 or anything else?

jrafanie:
> @jrafanie do these PRs have any dependencies? On [ManageIQ/manageiq#22854](https://github.com/ManageIQ/manageiq/pull/22854) or anything else?

@agrare We've been running with zeitwerk by default since https://github.com/ManageIQ/manageiq/pull/22488, so that's the dependency.

Technically, here is where we dropped support for the classic autoloader (merged about 1 month ago), making usage of these methods no longer needed: https://github.com/ManageIQ/manageiq/pull/22801

app/models/manageiq/providers/ovirt/inventory/collector.rb
jrafanie: We have similar/copied lines here: https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/40/files#diff-f2f5d4e887f15b36afacf467c592be51b1b1055f4aa04d5c553f77042620c671R2

Not sure what it's referring to but I can come back to it when I finish up the other PRs.  Maybe it was prior code that was removed and left orphaned. https://twitter.com/nzkoz/status/538892801941848064

jrafanie:
> > @jrafanie do these PRs have any dependencies? On [ManageIQ/manageiq#22854](https://github.com/ManageIQ/manageiq/pull/22854) or anything else?
> 
> @agrare We've been running with zeitwerk by default since [ManageIQ/manageiq#22488](https://github.com/ManageIQ/manageiq/pull/22488), so that's the dependency.
> 
> Technically, here is where we dropped support for the classic autoloader (merged about 1 month ago), making usage of these methods no longer needed: [ManageIQ/manageiq#22801](https://github.com/ManageIQ/manageiq/pull/22801)

Updated https://github.com/ManageIQ/manageiq/pull/22854 with this information.

PR #659, opened 2024-02-23
drop use of unuspported_reason_add

kbrock:
@agrare please take a look at this one. The patterns are different from the other providers

feels like a number of these could just delegate the reason to the failed supports check

---
part of:
- https://github.com/ManageIQ/manageiq/pull/22898
 
Deprecating `unsupported_reason_add`. Returning the string instead

Note, these are all the same:

```ruby
unsupported_reason_add(:feature, unsupported_reason(:feature2)) unless supports_feature2?
unsupported_reason_add(:feature, unsupported_reason(:feature2)) unless supports?(:feature2)
unsupported_reason(:feature2) unless supports?(:feature2)
unsupported_reason(:feature2)
```

### DIFF
--- a/app/models/manageiq/providers/ovirt/infra_manager.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager.rb
@@ -26,11 +26,11 @@ class ManageIQ::Providers::Ovirt::InfraManager < ManageIQ::Providers::InfraManag
   supports :admin_ui do
     # Link to oVirt Admin UI is supported for Engine version 4.1.8 or better.
     # See https://bugzilla.redhat.com/1512989 for details.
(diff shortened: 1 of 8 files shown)


kbrock:
update:
- fixed extra paren and missing `||`

kbrock:
update:
- fixed `action(:control)`
- swapped `remove_snapshot_denied_message` => `supports :remove_snapshot` per [`vm_snapshot_remove_one.rb`](https://github.com/kbrock/manageiq-ui-classic/blob/master/app/helpers/application_helper/button/vm_snapshot_remove_one.rb)

agrare (approved):

Issue #670, opened 2024-06-27
Targeted Refresh not queued for events


agrare:
When an Event is raised that should trigger a targeted refresh via [`EmsEvent#manager_refresh`](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event/automate.rb#L5-L13) no targeted refresh is queued.

The cause is due to the `EmsEvent#full_data` hash having symbol keys where the [`EventTargetParser#parse`](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/ovirt/infra_manager/event_target_parser.rb#L12-L22) method is expecting string keys.

PR #671, opened 2024-06-27
Fix EventTargetParser looking for string keys

agrare:
When we moved to kafka event payloads which commonly contained symbol keys were all being returned with string keys due to JSON being the intermediate format across kafka.  We decided to `deep_symbolize_keys` when processing the event payload https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_event_handler/runner.rb#L3 but RHV/Ovirt actually used string keys in their event payloads.

In order to support both kafka and MiqQueue we need to deep_symbolize_keys here

Dependents:
* https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/51

Fixes https://github.com/ManageIQ/manageiq-providers-ovirt/issues/670

### DIFF
--- a/app/models/manageiq/providers/ovirt/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ovirt/infra_manager/event_target_parser.rb
@@ -12,11 +12,11 @@ class ManageIQ::Providers::Ovirt::InfraManager::EventTargetParser
   def parse
     target_collection = InventoryRefresh::TargetCollection.new(:manager => ems_event.ext_management_system, :event => ems_event)
 
-    data = ems_event.full_data
+    data = ems_event.full_data.deep_symbolize_keys
 
-    add_vm_target(target_collection, data["vm"])             if data["vm"].present?
-    add_template_target(target_collection, data["template"]) if data["template"].present?
-    add_cluster_target(target_collection, data["cluster"])   if data["cluster"].present?
+    add_vm_target(target_collection, data[:vm])             if data[:vm].present?
+    add_template_target(target_collection, data[:template]) if data[:template].present?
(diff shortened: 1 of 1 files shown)


Fryguy:
I feel like this could be expensive with very deep payloads. Flipping this around, what if we deep_stringify_keys (or just assume it's all strings all the time and don't do aynthing)?  I guess I'm not understanding why everything has to be symbols.

agrare:
> I guess I'm not understanding why everything has to be symbols.

I tried to explain why in the PR description, I guess not very well.

Basically when using kafka we deep symbolize keys on the payload so this has to use symbols not strings.  In order to support the MiqQueue which would come in as strings (or existing events in the database) we have to be consistent with what the payload looks like.  We can change the event parser to pass symbols instead of strings and have a data migration to change existing ems_events but this is targeting backport to radjabov.

Fryguy:
Ok, yeah, so this is a tactical fix.   Ideally we'd switch to strings, but the reason for the original code's symbolizing was because all of EmsEvent and underlying layers expects symbols.  Unfortunate, but understandable.

agrare:
Yeah in a perfect world we'd standardize on JSON / string-keys everywhere but there was _a lot_ of code which would have to change and be re-tested.  As we move more towards kafka this would be definitely worthwhile though.

Fryguy:
Backported to `radjabov` in commit f38d1615f83272edd4e0caf901a1a1e018287e37.

```text
commit f38d1615f83272edd4e0caf901a1a1e018287e37
Date:   Fri Jun 28 13:24:12 2024 -0400

    Merge pull request #671 from agrare/fix_event_target_parser_event_string_keys
    
    Fix EventTargetParser looking for string keys
    
    (cherry picked from commit c7ce32d484940dd566fd1411a1f1ca39e4a10327)
```

PR #676, opened 2024-09-10
Update ovirt-engine-sdk to 4.6

kbrock:
Updating gem because it fixes some issues with ruby 3.3 building

Gem released Feb 1, 2024

I have no idea if I'm breaking anything here.
Just trying to make ruby 3.2 compile

### DIFF
--- a/manageiq-providers-ovirt.gemspec
+++ b/manageiq-providers-ovirt.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "ovirt-engine-sdk", "~>4.4.0"
+  spec.add_dependency "ovirt-engine-sdk", "~>4.6.0"
   spec.add_dependency "ovirt_metrics", "~>3.2"
 
   spec.add_development_dependency "manageiq-style"


kbrock:
update:
- changed from 4.6.1 to 4.6.0 (I misread ruby gems)

Fryguy:
Strange that we set ovirt_engine_sdk at all in red_hat_virtualization since it inherits from this gem anyway.

kbrock:
@Fryguy cool. I'll drop and see what happens

Fryguy:
The inheritance is kind of implied, but I don't see why we couldn't add a line to the gemspec to introduce it.

kbrock:
@miq-bot cross_repo_test #676 including https://github.com/ManageIQ/ovirt/pull/118

agrare:
@kbrock I see you included the "ovirt" gem in cross-repo, I don't think it is actually used any longer after 44446d6315efc0d63488d14fd5ecf40f80078ae1

kbrock:
```
Bundler could not find compatible versions for gem "ovirt-engine-sdk":
  In Gemfile:
    manageiq-providers-ovirt was resolved to 0.1.0, which depends on
      ovirt-engine-sdk (~> 4.6.0)

manageiq-providers-red_hat_virtualization was resolved to 0.1.0, which
depends on
      ovirt-engine-sdk (~> 4.4.0)
```

I updated red_hat_virtualization to depend upon this gem
and no longer specify `ovirt-engine-sdk`.

So I think this will be resolved once that is merged

agrare (approved):

Fryguy (approved):
LGTM, but I'd like to see cross-repo go green, ideally with red_hat_virtualization in there.  @kbrock can you update the cross-repo test to drop ovirt and add manageiq-providers-red_hat_virtualization?

PR #679, opened 2024-10-31
Delegate refresh_ems class method to parent infra manager

nasark:
Reasoning here: https://github.com/ManageIQ/manageiq-providers-azure/pull/564#issuecomment-2445034651

@miq-bot assign @agrare
@miq-bot add_label bug, radjabov/yes?
@miq-bot add_reviewer @agrare


agrare:
The Ovirt NetworkManager has its own RefreshWorker so we don't want to be delegating to the InfraManager

PR #685, opened 2025-01-27
Move default filtered event names to Settings

agrare:
https://github.com/ManageIQ/manageiq/issues/19707

### DIFF
--- a/app/models/manageiq/providers/ovirt/network_manager.rb
+++ b/app/models/manageiq/providers/ovirt/network_manager.rb
@@ -33,11 +33,7 @@ class ManageIQ::Providers::Ovirt::NetworkManager < ManageIQ::Providers::NetworkM
   end
 
   def self.default_blacklisted_event_names
-    %w(
-      scheduler.run_instance.start
-      scheduler.run_instance.scheduled
-      scheduler.run_instance.end
-    )
+    Settings.ems["ems_#{ems_type}"].blacklisted_event_names
   end
 
   def supported_auth_types
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -10,7 +10,13 @@
     :service:
       :read_timeout: 1.hour
       :open_timeout: 1.minute
-    :blacklisted_event_names: []
+    :blacklisted_event_names:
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/ovirt/network_manager.rb
```diff
@@ -33,11 +33,7 @@ def self.description
   end
 
   def self.default_blacklisted_event_names
-    %w(
-      scheduler.run_instance.start
-      scheduler.run_instance.scheduled
-      scheduler.run_instance.end
-    )
+    Settings.ems.ems_ovirt_network.blacklisted_event_names
```
Fryguy, line 36: If we did something like this, can we avoid creating new methods in the subclasses?

```suggestion
    Settings.ems["ems_#{ems_type}"].blacklisted_event_names
```

config/settings.yml
```diff
@@ -44,6 +50,11 @@
           - security_group.update.end
     :connection_manager:
       :purge_interval: 1.hour
+  :ems_ovirt_network:
```
Fryguy, line 53: Interesting - once we get rid of the BlacklistedEvent model and all that stuff, I wonder if worker_settings can handle provider subclassing in a way that merges this with the parent. Not required, but something we might want to consider if we want to avoid updating things in 2 places.

Fryguy (approved):
LGTM and I can merge as-is, but will wait on your reply to the above.

app/models/manageiq/providers/ovirt/network_manager.rb
agrare: Yeah we could do that we'd have to copy the settings over into all of the subclasses then I think

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-red_hat_virtualization/pull/59

agrare:
Okay CI fails here because the rhv PR isn't merged yet and it is testing seeding all providers, cross-repo is green with the rhv pr applied.

Fryguy (approved):

PR #690, opened 2025-06-06
Fix disks reconfigure via API

AndrianovAndrey:
MIQ-API converts string params into symbols, so we need to convert it back.


AndrianovAndrey:
@miq-bot add-label bug

AndrianovAndrey:
@agrare yes, thank you  for the quick reply, made a conversion up in build_config_spec, now it should be just fine.

agrare:
Rebased on master, fixed rubocop warnings, and fixed the network_spec method using symbols causing spec failures

agrare:
@Fryguy I have a commit in here now can you review?

